PR #33, opened 2017-10-21
Ember: Convert `rootElement` to non-caching property

Turbo87:
This makes the library compatible with the recent changes in `ember-test-helpers` where the `#ember-testing` element is recreated for each test instead of reusing the existing one.

/cc @rwjblue


rwjblue:
👍

PR #36, opened 2017-11-03
Add `doesNotIncludeText()` assertion

Zureka:
There appears to be a pattern of having a positive and a negative assertion for any given property/attribute. E.g. `isFocused` and `isNotFocused`, `hasAttribute` and `doesNotHaveAttribute`, and `hasClass` and `doesNotHaveClass`. 

With that being said, it appears that there doesn't appear to be a corresponding inverse for the `includesText` assertion. Until now!

### DIFF
--- a/lib/assertions.js
+++ b/lib/assertions.js
@@ -333,6 +333,8 @@ export default class DOMAssertions {
    * [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)
    * attribute.
    *
+   * **Aliases:** `doesNotContainText`, `doesNotHaveTextContaining`
+   *
    * @param {string} text
    * @param {string?} message
    *
@@ -352,6 +354,14 @@ export default class DOMAssertions {
     }
 
     this.pushResult({ result, actual, expected, message });
+  }
+
+  doesNotContainText(expected, message) {
+    this.doesNotIncludeText(expected, message);
+  }
+
+  doesNotHaveTextContaining(expected, message) {
+    this.doesNotIncludeText(expected, message);
   }
 
   /**
--- a/lib/assertions.js
+++ b/lib/assertions.js
@@ -328,6 +328,33 @@ export default class DOMAssertions {
   }
 
   /**
+   * Assert that the text of the [HTMLElement][] or an [HTMLElement][]
+   * matching the `selector` does not contain the given `text`, using the
+   * [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)
+   * attribute.
+   *
+   * @param {string} text
+   * @param {string?} message
+   *
+   * @example
+   * assert.dom('#title').doesNotIncludeText('Welcome');
+   */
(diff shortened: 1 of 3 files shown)


lib/assertions.js
```diff
@@ -328,6 +328,43 @@ export default class DOMAssertions {
...
+   * @example
+   * assert.dom('#title').doesNotIncludeText('Welcome');
+   */
+  doesNotIncludeText(text, message) {
+    let element = this.findTargetElement();
+    if (!element) return;
+
+    let result = element.textContent.indexOf(text) === -1;
+    let actual = element.textContent;
+    let expected = text;
```
Turbo87: `expected = text` does not make sense in this case, right?

Zureka: Good catch! Would you prefer the name `unexpected` or something else?

Turbo87: maybe something like `` `does not include "${text}"` ``?

Zureka: I can certainly update the phrasing of the message to `does not include "${text}"`, my question is actually regarding what is passed into `this.pushResult` at the end of this function. Would it be fine to say `this.pushResult({ results, actual, unexpected, message });`?

Turbo87: no, because the `pushResult` method explicitly requires an `expected` attribute (see api.qunitjs.com/assert/pushResult)

Zureka: That's what I figured. `this.pushResult({ result, actual, expected: unexpected, message});` would work but seems a little ironic to "expect the unexpected".

Zureka:
@Turbo87 Is there anything else that's needed for this to be merged?

Turbo87:
@Zureka my initial concern at https://github.com/simplabs/qunit-dom/pull/36#discussion_r149044417 is still valid. QUnit will show something as "expected" that was actually not expected.

lib/assertions.js
```diff
@@ -327,6 +327,47 @@ export default class DOMAssertions {
...
+    let result = element.textContent.indexOf(text) === -1;
+    let expected = `Element ${this.targetDescription} does not include text "${text}"`;
+    let actual = expected;
+
+    if(!result) {
+      actual = `Element ${this.targetDescription} includes text "${text}"`;
+    }
+
+    if (!message) {
+      message = `Expected element ${this.targetDescription} to not include text "${text}"`;
```
Turbo87: I think we can `message = expected` here like in a few of the other assertions

lib/assertions.js
```diff
@@ -327,6 +327,47 @@ export default class DOMAssertions {
...
+   */
+  doesNotIncludeText(text, message) {
+    let element = this.findTargetElement();
+    if (!element) return;
+
+    let result = element.textContent.indexOf(text) === -1;
+    let expected = `Element ${this.targetDescription} does not include text "${text}"`;
+    let actual = expected;
+
+    if(!result) {
```
Turbo87: there is a space missing here after the `if`

Turbo87:
Thanks @Zureka!

PR #40, opened 2017-11-28
[Ember] Adjust `rootElement`

### LINKED ISSUE #39: [Ember] different root element
I have quite a few failing tests after applying the codemod, for tests that were using the `:first-child` selector. This is due to the fact that in normal Ember tests the first child will be the component under test, while here it is another wrapper element inside `#ember-testing`. This is similar to the same issue in `ember-native-dom-helpers`: https://github.com/cibernox/ember-native-dom-helpers/issues/63.

I used [this](https://github.com/kaliber5/ember-bootstrap/blob/master/tests/test-helper.js#L16) to fix it for my use, the same selector could be used in your vendor script. Can submit a PR if you agree!?

simonihmig:
Fixes #39

### DIFF
--- a/vendor/overwrite-qunit-dom-root-element.js
+++ b/vendor/overwrite-qunit-dom-root-element.js
@@ -1,6 +1,6 @@
 Object.defineProperty(QUnit.assert.dom, 'rootElement', {
   get: function() {
-    return document.querySelector('#ember-testing');
+    return document.querySelector('#ember-testing > .ember-view') || document.querySelector('#ember-testing');
   },
   enumerable: true,
   configurable: true,
--- a/tests/integration/root-element-test.js
+++ b/tests/integration/root-element-test.js
@@ -1,0 +1,12 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('root-element', 'Integration | root element', {
+  integration: true
+});
+
+test('it has correct root element', function(assert) {
+  this.render(hbs`<div class="first-child"></div>`);
+
+  assert.dom(':first-child').hasClass('first-child', ':first-child returns the first element');
+});


tests/acceptance/qunit-dom-test.js
```diff
@@ -9,10 +9,10 @@ test('qunit-dom assertions are available', function(assert) {
   assert.ok(assert.dom, 'assert.dom is available');
   assert.ok(assert.dom('.foo').includesText, 'assert.dom(...).includesText is available');
 
-  assert.dom('#qunit').doesNotExist('rootElement is set to #ember-testing-container');
-
   visit('/');
   andThen(() => {
+    assert.dom('#qunit').doesNotExist('rootElement is set to #ember-testing-container');
```
simonihmig: Had to move this, because the `.ember-view` seems to be rendered after the `visit` call. Before that `document.querySelector('#ember-testing > .ember-view')` will return `null`...

simonihmig:
Fyi, https://github.com/cibernox/ember-native-dom-helpers/pull/121 has been merged...

vendor/overwrite-qunit-dom-root-element.js
```diff
@@ -1,6 +1,6 @@
 Object.defineProperty(QUnit.assert.dom, 'rootElement', {
   get: function() {
-    return document.querySelector('#ember-testing');
+    return document.querySelector('#ember-testing > .ember-view');
```
Turbo87: Can we make this fallback to the old behavior if the new one can't be found?

tests/acceptance/qunit-dom-test.js
Turbo87: Shouldn't be needed with the proposal below.

vendor/overwrite-qunit-dom-root-element.js
simonihmig: You mean something like `return document.querySelector('#ember-testing > .ember-view') || document.querySelector('#ember-testing')`?

Turbo87: yep, exactly

simonihmig: Done! Also reverted the changes in the other test, as suggested!

Turbo87:
Thanks! 👍

simonihmig:
Thanks for merging! Would you be able to cut a release with this? Would like to integrate this into ember-bootstrap, where I had some problems because of this issue...

Turbo87:
yep, just published `v0.4.0`

PR #50, opened 2018-02-02
Collapse non-breaking whitespace.

CvX:
`&nbsp;` entity wasn't being collapsed, so HTML like `<div>10&nbsp;users online</div>` couldn't be matched with `assert.dom('div').hasText(/10 users/)`

This PR replaces `/ +/` regex with `/\s+/`.
Test included.

### DIFF
--- a/lib/helpers/collapse-whitespace.js
+++ b/lib/helpers/collapse-whitespace.js
@@ -1,7 +1,7 @@
 export default function collapseWhitespace(string) {
   return string
     .replace(/[\t\r\n]/g, ' ')
-    .replace(/ +/g, ' ')
+    .replace(/\s+/g, ' ')
     .replace(/^ /, '')
     .replace(/ $/, '');
 }
--- a/lib/helpers/collapse-whitespace.test.js
+++ b/lib/helpers/collapse-whitespace.test.js
@@ -2,6 +2,9 @@
 
 import collapseWhitespace from './collapse-whitespace';
 
+let testElement = document.createElement('div');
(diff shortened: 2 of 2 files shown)


Turbo87:
I'm not convinced that we should be collapsing `&nbsp;` characters. If you use HTML like `a&nbsp;&nbsp;&nbsp;b` it will be displayed as  `a   b`, not `a b`, so collapsing whitespace in this case seems wrong to me 🤔

CvX:
That's a good point. How about this: collapse whitespace as before, but after that replace each `/\s/` with a space character. i.e. https://github.com/CvX/qunit-dom/commit/f7bcef277dc9c21b6322efc7fba1ee0cc0565e3e

Turbo87:
but... why? the whitespace collapsing that we have now should be similar to what is defined in the standard, so I'm not sure why we should change it?

CvX:
It's just the matter of ergonomics. For example, to match the `<span>10&nbsp;users</span>\nonline` you can't use `/10 users online/` but instead: /10\susers\s+online/.

It requires to think ahead about the structure of the DOM and whitespace in the template when writing the test. I'd argue that it shouldn't be the case when just testing if given element has some text.

Either way, it's not a big deal, but can be surprising when the test fails and both expected and actual strings look the same. 😅

Turbo87:
you could use `.hasText('10\xa0users online')` which makes it more obvious that there is a nbsp involved

CvX:
Alright, I'll stick with `\s` and `\s+` in my tests. I don't find UTF hex more readable. 😅

PR #54, opened 2018-02-22
Add isVisible() and isNotVisible() assertions

patocallaghan:
Fixes https://github.com/simplabs/qunit-dom/issues/52

Adds `isVisible` and `isNotVisible` assertion helpers.

Visibility is modelled off of jQuery's `:visible` logic in jQuery 3.0 https://github.com/jquery/jquery/blob/4a2bcc27f9c3ee24b3effac0fbe1285d1ee23cc5/src/css/hiddenVisibleSelectors.js#L11-L13

From https://api.jquery.com/hidden-selector/, an element is considered "hidden" if it follows any of the following criteria

> They have a CSS display value of none.
They are form elements with type="hidden".
Their width and height are explicitly set to 0.
An ancestor element is hidden, so the element is not shown on the page.
Elements with visibility: hidden or opacity: 0 are considered to be visible, since they still consume space in the layout.
elements will be considered :hidden if they don't have any layout boxes. For example, br elements and inline elements with no content will not be selected by the :hidden selector.

As mentioned in the issue you may decide not to merge this. If you do, I can then update the README information.

### DIFF
--- a/lib/assertions.js
+++ b/lib/assertions.js
@@ -2,6 +2,7 @@ import exists from './assertions/exists';
 import focused from './assertions/focused';
 import notFocused from './assertions/not-focused';
 
+import visible from './helpers/visible';
 import elementToString from './helpers/element-to-string';
 import collapseWhitespace from './helpers/collapse-whitespace';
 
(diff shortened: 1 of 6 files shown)


lib/helpers/visible.js
```diff
@@ -0,0 +1,6 @@
+// Visible logic based on jQuery's
+// https://github.com/jquery/jquery/blob/4a2bcc27f9c3ee24b3effac0fbe1285d1ee23cc5/src/css/hiddenVisibleSelectors.js#L11-L13
+
+export default function visible(el) {
+  return !!(el.offsetWidth || el.offsetHeight || el.getClientRects().length);
```
Turbo87: could make this check a little more explicit? assuming I understand the logic correctly, maybe something like this:

```js
return el.offsetWidth !== 0 || el.offsetHeight !== 0 || el.getClientRects().length !== 0
```

lib/assertions.js
```diff
@@ -462,6 +463,58 @@ export default class DOMAssertions {
     this.hasNoValue(message);
   }
 
+  /**
+   * Assert an [HTMLElement][] matching the `selector` is visible.
```
Turbo87: can you add to the documentation what makes an element "visible"? for example if the browser is scrolled to a different part of the page does that mean that elements are not visible then?

lib/helpers/visible.js
scalvert: This seems like a good change, though the above is literally what jQuery uses (though that isn't any endorsement for using that code).

scalvert:
I tried to push some of these fixes to @patocallaghan's fork, but no dice. Guess we'll have to wait.

scalvert:
So one thing that needs to be addressed here is that this code will fail if an element is not rendered yet, which may still be considered not visible. 

In jQuery, you always get _an_ element back. So, the visible check works well because if jQuery doesn't find the element, it returns a default jQuery object which can be evaluated. 

With using `querySelector`, if an element is not found (not rendered yet) it will return `null`.  Ember's test-helpers use `find`, which under the covers use `querySelector`. This means that when passing in an element that was the result of `find` we could end up passing `null` in. This will currently error in the above.

To fix, we should first test for the element's existence:

```js
return !!el && (el.offsetWidth !== 0 || el.offsetHeight !== 0 || el.getClientRects().length !== 0);
```

patocallaghan:
Apologies I hope to get to addressing the comments and finishing this in the next few days

scalvert:
Let me know what i can do to help.

patocallaghan:
Closing this PR as #67 supersedes it

Issue #56, opened 2018-03-11
Add isRequired() and isNotRequired() assertions


Turbo87:
This assertion should check if the `required` property of e.g. an `<input>` element is `true` or not.

#### Example: 
```js
assert.dom('.username-field').isRequired();
assert.dom('.coupon-code').isNotRequired();
```

scalvert:
Also, the attribute being present also represents `true`, so the assertion should take that into consideration too.

billybonks:
@Turbo87 guess you can mark this as complete

PR #65, opened 2018-03-16
Add isRequired and isNotRequired assertions

scalvert:
Implements #56 

Fills in the API for
- `isRequired`
- `isNotRequired`

TODO:
- [x] implement `isRequired`
- [x] implement `isNotRequired`
- [x] write tests
- [x] write docs
- [x] regenerate docs [rendered](https://github.com/scalvert/qunit-dom/blob/3470f2e776206017ad6fb080256a20a43147ecae/API.md)


lib/assertions/is-not-required.js
```diff
@@ -0,0 +1,16 @@
+import elementToString from '../helpers/element-to-string';
+
+export default function notRequired(message) {
+  let element = this.findTargetElement();
+  if (!element) return;
+
+  let result = element.required === false;
+  let actual = element.required === true ? 'required' : 'not required';
```
Turbo87: I think you can just reuse the `result` variable here instead of comparing the property again

scalvert: Done. Also updated `is-required.js` to do the same. FWIW, this is what `is-checked` does, so we may want to update that too separately.

Turbo87: yeah, agreed. I merged it and planned to do a small cleanup round later :)

lib/assertions/is-not-required.js
```diff
@@ -0,0 +1,16 @@
+import elementToString from '../helpers/element-to-string';
+
+export default function notRequired(message) {
+  let element = this.findTargetElement();
+  if (!element) return;
+
+  let result = element.required === false;
```
Turbo87: what happens if the element is not an `<input>` element? would `required` be `undefined` then and fail this check?

scalvert: Good catch. I've added a conditional, which will throw if an element is passed that is not one of `HTMLInputElement`, `HTMLTextAreaElement`, or `HTMLSelectElement`.

scalvert: And a test...

scalvert:
@Turbo87 anything else I need to do to push this in?

Turbo87:
> anything else I need to do to push this in?

remind me about it 😜

PR #66, opened 2018-03-23
Add table of contents to API docs

ddoria921:
Created table of contents with links to each section


Turbo87:
@ddoria921 thanks! unfortunately there is a small problem: `api.md` is automatically generated and once we regenerate it the ToC will be gone again 😱 

we might however be able to take advantage of something like https://github.com/emberjs/ember-test-helpers/blob/master/documentation.yml which then should be able to automatically generate a ToC from that (see https://github.com/emberjs/ember-test-helpers/blob/master/API.md#table-of-contents).

ddoria921:
@Turbo87 I configured the `docs` script in the `package.json` to use the `build` command instead of the `readme` command. Then, I added configuration file. In the configuration I specified to include a `assert.dom()` section at the top and I copied over the example used in the existing API doc. 

Check it out and let me know what you think. The end result looks a little different than how it is currently formatted, but it's the same information.

Turbo87:
thanks :)

PR #70, opened 2018-03-27
Added `hasEmptyAttribute`

pizza:
👋 In an increasing number of cases in the codebase I work on mostly, we are starting to check if a given DOM element has an attribute, and we're big fans of adding messages whenever possible to the assertion.

`hasAttribute` is great, but (obviously) only accepts a `message` argument if it's the third parameter in `hasAttribute(name, value, message)`.

Given that we increasingly need to check for empty attributes (e.g.: custom data attributes), it would be great if instead of
`hasAttribute('myName', '', 'this DOM has myName')`
we could have
`hasEmptyAttribute('myName', 'this DOM has myName')`

--

When working on this change, I based my code on `hasAnyText`, so `hasEmptyAttribute` is almost an alias of `hasAttribute` and gets its own test file.

Let me know what you think.

### DIFF
--- a/lib/__tests__/has-empty-attribute.js
+++ b/lib/__tests__/has-empty-attribute.js
@@ -1,0 +1,89 @@
+/* eslint-env jest */
+
+import TestAssertions from "../helpers/test-assertions";
+
+describe('assert.dom(...).hasEmptyAttribute()', () => {
+  let assert;
+
+  beforeEach(() => {
+    assert = new TestAssertions();
+
+    document.body.innerHTML = '<button class="share" checked></button>';
+  });
+
+  test('succeeds for attribute without value', () => {
+    assert.dom('.share').hasAttribute('checked', '');
+    assert.dom('.share').hasEmptyAttribute('checked');
+    assert.dom(document.querySelector('.share')).hasEmptyAttribute('checked');
+
+    expect(assert.results).toEqual([{
+      actual: 'Element .share has attribute "checked" with value ""',
+      expected: 'Element .share has attribute "checked" with value ""',
+      message: 'Element .share has attribute "checked" with value ""',
+      result: true,
+    }, {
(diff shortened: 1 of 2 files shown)


Turbo87:
> it would be great if instead of
> hasAttribute('myName', '', 'this DOM has myName')
> we could have
> hasEmptyAttribute('myName', 'this DOM has myName')

I'm not sure I agree. Using `hasEmptyAttribute` is actually one additional character, and it has the potential for confusion since I could imagine people thinking that a missing attribute is also considered and "empty" attribute. 🤔

pizza:
I don't think that an empty attribute is the same as a missing attribute. There's already `doesNotHaveAttribute` for the latter. But I'm open to a better name.

Two more use cases where I think that `hasEmptyAttribute` can be useful are:
- when used with [ember-test-selectors](https://github.com/simplabs/ember-test-selectors): `hasEmptyAttribute('data-test-active', 'x is active')`
- when checking for `checked` or `disabled` or other boolean attributes: `hasEmptyAttribute('checked', 'x is checked')`.

Turbo87:
> `hasEmptyAttribute('data-test-active', 'x is active')`

I don't think that's what you want to test. It looks like what you want to test is whether the attribute exists at all independent from the value, which is exactly what `hasAttribute('data-test-active')` does.

> when checking for checked or disabled or other boolean attributes: hasEmptyAttribute('checked', 'x is checked').

Same as above. The HTML spec does require a specific attribute value for these things as long as the attribute exists, so e.g. `hasAttribute('checked')` is sufficient. For `checked` and `disabled` we should also not check the attributes, but check the properties instead, and we now have dedicated assertions for those.

IMHO the proposed `hasEmptyAttribute()` methods adds little value over using the plain `hasAttribute()` method with an empty string, but it does increase the overall size of the API surface. Thanks you for the proposal and for working on this, but due to the reasons mentioned above I will close the PR now.

PR #76, opened 2018-04-10
Reduce specificity of Element check in findTargetElement.

adriancooney:
Fixes https://github.com/simplabs/qunit-dom/issues/75.

I wasn't entirely sure where to put a test case for this so I said I'd wait for feedback from someone with some context. Has anyone any suggestions on where to put a test for this?

### DIFF
--- a/lib/assertions.js
+++ b/lib/assertions.js
@@ -670,7 +670,7 @@ export default class DOMAssertions {
 
       return el;
 
-    } else if (this.target instanceof HTMLElement) {
+    } else if (this.target instanceof Element) {
       return this.target;
 
     } else {


adriancooney:
@Turbo87 funnily enough, that's exactly where I wanted to put the test in the first place myself and I found that `exists` doesn't use the `findTargetElement` method. I could throw one into `hasAttribute` tests?

Turbo87:
> and I found that exists doesn't use the findTargetElement method.

whoops, you're right :D

> I could throw one into hasAttribute tests?

yeah, seems good to me :)

adriancooney:
@Turbo87 test added!

PR #82, opened 2018-04-15
Strip non-breaking and zero-width spaces in `collapseWhitespace` helper

patocallaghan:
We noticed when migrating some assertions in our tests from an internal whitespace helper to qunit-dom's `hasText` that we had some failures. It was showing a mismatch due to `collapseWhitespace` not collapsing non-breaking spaces and zero-width spaces.

### DIFF
--- a/lib/helpers/collapse-whitespace.js
+++ b/lib/helpers/collapse-whitespace.js
@@ -1,7 +1,7 @@
 export default function collapseWhitespace(string) {
   return string
-    .replace(/[\t\r\n]/g, ' ')
-    .replace(/ +/g, ' ')
+    .replace(/\s+/g, ' ')
+    .replace(/\u200b+/g, '')
     .replace(/^ /, '')
     .replace(/ $/, '');
 }
--- a/lib/helpers/collapse-whitespace.test.js
+++ b/lib/helpers/collapse-whitespace.test.js
@@ -10,6 +10,7 @@ const TESTS = [
   [' a b c ', 'a b c'],
   [' a\r\n b c ', 'a b c'],
   ['\n    foo equals\n    bar\n  ', 'foo equals bar'],
+  ['a\xa0b\u200bc', 'a bc'],
 ];
 
 TESTS.forEach(it => {


lib/helpers/collapse-whitespace.js
```diff
@@ -1,7 +1,7 @@
 export default function collapseWhitespace(string) {
   return string
-    .replace(/[\t\r\n]/g, ' ')
-    .replace(/ +/g, ' ')
+    .replace(/\s+/g, ' ')
+    .replace(/\u200b+/g, '')
     .replace(/^ /, '')
```
patocallaghan: Is there a particular reason we don't use `.trim()` here? It's ES5 so pretty much [all browsers these days support it](http://kangax.github.io/compat-table/es5/#test-String_properties_and_methods_String.prototype.trim_a_href=_https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim_title=_MDN_documentation_img_src=_../mdn.png_alt=_MDN_(Mozilla_Development_Network)_logo_width=_15_height=_13_/_/a_nbsp;)

CvX:
@patocallaghan see discussion in #50 😁

patocallaghan:
@CvX Oh interesting, on reflection I agree with that, we shouldn't strip non-breaking spaces from within actual content.

On the other hand, I do think we should strip non-breaking spaces from the start and end of a string as that's exactly how [ES5's `trim()` works](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim). That was more the use case that was affecting my tests.

Also thinking about it, removing the zero-width space doesn't seem appropriate here either, so I'll probably remove that.

CvX:
I'd agree with using `trim()` instead of replacing space characters at the beginning and end, but then again I'm biased against matching any whitespace. 😉

patocallaghan:
@Turbo87 On consideration I think you are right with regards zero-width space and in-content html characters.

On the other hand, would you consider any sort any variant of `hasText` which would use `.trim()` to strip whitespace from the beginning and end of the string, possible as an option? For example, [chai-dom](https://github.com/nathanboktae/chai-dom#texttext) offers this behaviour on `.text` via a flag.

Turbo87:
> On the other hand, would you consider any sort any variant of hasText which would use .trim() to strip whitespace from the beginning and end of the string, possible as an option? For example, chai-dom offers this behaviour on .text via a flag.

my goal is to keep the API as simple as possible, and I personally haven't yet ran into any situation where I would have needed that. can you give a few real-life examples of where this would be useful?

patocallaghan:
@Turbo87 sure I understand. I'll see if I can come up with real-life examples 👍

patocallaghan:
So I managed to convert all of our test cases and it wasn't as much of an issue as I initially thought. I managed to use `containsText` in the cases where `.trim()` may have been an issue. Closing this PR.

PR #89, opened 2018-05-22
Improve visibility detection

Turbo87:
Previously the algorithm only checked if client rects existed, but it did not check if those rects had an actual size. We now check the returned array from `getClientRects()` if the rects do have a non-zero size.


patocallaghan:
oh, nice 👍 So I guess this is a bug in jQuery too? 😄

Turbo87:
I would assume so 🤔

Issue #84, opened 2018-04-19
`Element <target> exists` error message when it does not.


adriancooney:
So while using the `isNotVisible`, I tested on a selector that did not return any elements and got the following error:

```
assert.dom('.my_selector').isNotVisible();
// Throws: Element .my_selector exists
```

This message to begin with is a little cryptic because it doesn't mention visibility anywhere (which could be something we should look into). Secondly, elements matching the `.my_selector` *don't* exist! Exactly the opposite of what the error message suggests. This could be considered a bug, right? I think the offending code [lies in `findTargetElement`](https://github.com/simplabs/qunit-dom/blob/master/lib/assertions.js#L679-L700) with some interesting conditions:

```js
  findTargetElement() {
    if (this.target === null) {
      let message = `Element <unknown> exists`;
      this.pushResult({ message, result: false });
      return null;
    }

    if (typeof this.target === 'string') {
      let el = this.rootElement.querySelector(this.target);

      if (el === null) {
        let message = `Element ${this.target || '<unknown>'} exists`;
        this.pushResult({ message, result: false });
      }

      return el;

    } else if (this.target instanceof Element) {
      return this.target;

    } else {
      throw new TypeError(`Unexpected Parameter: ${this.target}`)
    }
  }
```

Am I missing something? Or any ideas on a fix? I can submit a PR if necessary.

scalvert:
I also found a bug in the `isVisible`/`isNotVisible` assertions. The current implementation of `findTargetElement` is intertwined with `pushResult` calls. This causes issues when an element doesn't yet exist in the DOM, as  it triggers this:

```js
      if (el === null) {
        let message = `Element ${this.target || '<unknown>'} exists`;
        this.pushResult({ message, result: false });
      }
```
as @adriancooney suggests. 

There's two things here:
1. I'm not sure how the tests pass for this right now
2. I believe the `findTargetElement` function needs to be decomposed somewhat to disconnect the finding of elements from `pushResult` calls.

I'm going to take a stab at fixing this issue, as the visibility checks aren't working properly right now.

cc: @rwjblue

scalvert:
With the above PR merged, does this address your original concern @adriancooney?

adriancooney:
That would have cleared it up for me @scalvert, thank you!

PR #91, opened 2018-05-30
Fixing is-visible/is-not-visible assertions to correctly account for missing elements

scalvert:
This PR is attempting to do two main things:

1. Fixing the logic in `is-visible`/`is-not-visible` to account for when elements are missing in the DOM
    - these assertions would trigger the `Element <el> exists` assertion that was triggered when an element was not found in the DOM, regardless of correctness
    - `is-visible` should *pass* in the case where the element _exists_ *and* is visible
    - `is-visible` should *fail* in the case where the element _doesn't exist_ *or* is not visible
    - `is-not-visible` should *pass* in the case where the element _doesn't exist_ *or* is not visible
    - `is-not-visible` should *fail* in the case where the element _exists_ *and* is visible
1. Updating the assertion message when an element is not present but is expected to be so
    - the *old* assertion was `Element <el|unknown> exists`, which was mildly confusing
    - the *new* assertion message is `Element <el|unknown> should exist`.

This PR in part addresses #84.


scalvert:
@Turbo87 - I think we may want to address this sooner rather than later, as the assertions are currently broken.

.gitignore
```diff
@@ -21,3 +21,4 @@ testem.log
 .node_modules.ember-try/
 bower.json.ember-try
 package.json.ember-try
+.vscode/
```
rwjblue: In general, editor specific ignores like this should be added to the [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

lib/assertions.js
```diff
@@ -673,28 +673,42 @@ export default class DOMAssertions {
...
+    return el;
+  }
+
+  /**
+   * Finds a valid HTMLElement from target
+   * @private
+   * @returns a valid HTMLElement, or null
+   * @throws TypeError will be thrown if target is an unrecognized type
+   */
+  findElement() {
```
rwjblue: Since this is private, should it be underscored (I'm not sure what the norm is in qunit-dom)?

Also, is there a name that better describes the split between this method and `findTargetElement`?  Maybe `findTargetElementWithoutAssertion` (yeah, I know its pretty long)?

scalvert:
@rwjblue added the comment to cross-reference the tests in their various locations.

lib/assertions.js
scalvert: Ya I went with no underscore as this repo had set the precedent of no underscore. FWIW, I prefer underscore...

rwjblue:
Also, I do think that this fully addresses #84. It both tweaks the assertion of `findTargetElement` _and_ makes `assert.dom('.foo').isNotVisible()` actually function properly (e.g. it passes when `.foo` is not present).

lib/assertions.js
scalvert: I can rename the new method, but it does seem a bit backwards. The `findElement` method is the simple case, and `findTargetElement` is the method that's layering on functionality. As such, I would likely want to rename `findTargetElement` to something like `findElementWithAssertion` rather than the opposite (renaming the smaller, more well defined method).

.gitignore
scalvert: Will add.

scalvert:
Ya I was more speaking that this PR does _more_ than what's needed in #84, hence partial. English is hard...

Turbo87:
@scalvert we're pretty busy with JSConf EU prep at the moment. It will likely take until next week before I have time for a proper review. On first glance it seems that the PR is trying to solve multiple things in parallel. Can you extract the `Element <el|unknown> should exist` change into a separate PR?

scalvert:
Yep. I definitely conflated the changes in this PR. Will extract.

PR #92, opened 2018-05-31
Changes element assertion from 'exists' to 'should exist'

scalvert:
Updating the assertion message when an element is not present but is expected to be so
- the old assertion was Element <el|unknown> exists, which was mildly confusing as it was often accompanied with a test failure, but asserted an error in the positive voice
- the new assertion message is Element <el|unknown> should exist, which is an assertion as to the expected state

This PR addresses #84.

### DIFF
--- a/lib/assertions/exists.js
+++ b/lib/assertions/exists.js
@@ -41,7 +41,7 @@ export default function exists(options, message) {
 
 function format(selector, num) {
   if (num === undefined || num === null) {
-    return `Element ${selector} exists`;
+    return `Element ${selector} should exist`;
   } else if (num === 0) {
     return `Element ${selector} does not exist`;
   } else if (num === 1) {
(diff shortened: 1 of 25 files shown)


lib/assertions/exists.js
```diff
@@ -41,7 +41,7 @@ export default function exists(options, message) {
 
 function format(selector, num) {
   if (num === undefined || num === null) {
-    return `Element ${selector} exists`;
+    return `Element ${selector} should exist`;
```
Turbo87: I'm not sure this one should be changed too 🤔

scalvert: Ya I wasn't sure. Could be convinced either way. I could see the argument for an `exists` assertion saying something **exists** :P

scalvert: I've reverted this change.

scalvert:
O_o

scalvert:
:+1:

PR #93, opened 2018-05-31
Let `isNotVisible()` pass if the element does not exist

scalvert:
Fixing the logic in is-visible/is-not-visible to account for when elements are missing in the DOM.
These assertions would trigger the Element <el> exists assertion that was triggered when an element was not found in the DOM, regardless of correctness
- `is-visible` should **pass** in the case where the element **exists** and is visible
- `is-visible` should **fail** in the case where the element **doesn't exist** or is not visible
- `is-not-visible` should **pass** in the case where the element **doesn't exist** or is not visible
- `is-not-visible` should **fail** in the case where the element **exists** and is visible

### DIFF
--- a/lib/assertions.js
+++ b/lib/assertions.js
@@ -673,21 +673,37 @@ export default class DOMAssertions {
   }
 
   /**
+   * Finds a valid HTMLElement from target, or pushes a failing assertion if a valid
+   * element is not found.
    * @private
+   * @returns a valid HTMLElement, or null
    */
   findTargetElement() {
-    if (this.target === null) {
-      let message = `Element <unknown> should exist`;
+    let el = this.findElement();
+
+    if (el === null) {
+      let message = `Element ${this.target || '<unknown>'} should exist`;
       this.pushResult({ message, result: false });
       return null;
     }
 
-    if (typeof this.target === 'string') {
+    return el;
+  }
+
+  /**
+   * Finds a valid HTMLElement from target
+   * @private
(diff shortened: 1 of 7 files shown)


scalvert:
O_o

Turbo87:
looks like this has conflicts now... 😞 

I tend to agree that this logic is probably better. I'll review the code changes hopefully tomorrow.

scalvert:
I’ll fix the conflicts.

scalvert:
@Turbo87 fixed!

lib/assertions.js
```diff
@@ -226,34 +226,54 @@ export default class DOMAssertions {
...
+      let actual =
+        actualValue === null
+          ? `Element ${
+              this.targetDescription
+            } does not have attribute "${name}"`
+          : `Element ${
+              this.targetDescription
+            } has attribute "${name}" with value ${JSON.stringify(
+              actualValue
+            )}`;
```
Turbo87: @scalvert this looks like unrelated changes 🤔 

I'm assuming you used prettier here, but I'm not sure if the result in this case is actually prettier than before... 😄

scalvert: Damn. It’s vscode. I format on save for some repos and not for others. It’s maddening. Will fix.

scalvert:
@Turbo87 ok I've managed to wrangle this PR into a good state. Can you take another look?

Turbo87:
Merci! :)

PR #99, opened 2018-07-10
Add support for `aria-disabled` to `isDisabled()` assertion

billybonks:
Added support for aria-disabled

allows supporting this property on generic containers

https://codepen.io/billybonks/pen/YjzLJN?editors=1010 

take a look at console.log :)

### DIFF
--- a/lib/__tests__/is-disabled.ts
+++ b/lib/__tests__/is-disabled.ts
@@ -133,6 +133,6 @@ describe('assert.dom(...).isDisabled()', () => {
     expect(() => assert.dom(undefined).isDisabled()).toThrow('Unexpected Parameter: undefined');
     expect(() => assert.dom({}).isDisabled()).toThrow('Unexpected Parameter: [object Object]');
     expect(() => assert.dom(document).isDisabled()).toThrow('Unexpected Parameter: [object Document]');
-    expect(() => assert.dom(document.createElement('div')).isDisabled()).toThrow('Unexpected Element Type: [object HTMLDivElement]');
+    expect(() => assert.dom(document.createElement('div')).isDisabled()).toThrow('Generic Element Type: [object HTMLDivElement] does not use aria-disabled attribut');
   });
 });
--- a/lib/__tests__/is-disabled.ts
+++ b/lib/__tests__/is-disabled.ts
@@ -91,9 +91,9 @@ describe('assert.dom(...).isDisabled()', () => {
             assert.dom('#child').isDisabled();
 
             expect(assert.results).toEqual([{
-              actual: 'Element #child is disabled',
-              expected: 'Element #child is disabled',
-              message: 'Element #child is disabled',
+              actual: 'Parent of Element #child is disabled',
+              expected: 'Parent of Element #child is disabled',
+              message: 'Parent of Element #child is disabled',
               result: true,
             }]);
           });
--- a/lib/__tests__/is-disabled.ts
+++ b/lib/__tests__/is-disabled.ts
@@ -82,6 +82,33 @@ describe('assert.dom(...).isDisabled()', () => {
       document.body.innerHTML = '<input type="text" disabled>';
     });
 
+    describe('when using a generic container', () => {
+
+      test('succeeds if element is disabled', () => {
+        document.body.innerHTML = '<div aria-disabled="true"></div>';
+        assert.dom('div').isDisabled();
+
+        expect(assert.results).toEqual([{
+          actual: 'Element div is disabled',
+          expected: 'Element div is disabled',
+          message: 'Element div is disabled',
+          result: true,
+        }]);
+      });
+
+      test('fails if element is not disabled', () => {
+        document.body.innerHTML = '<div aria-disabled="false"></div>';
+        assert.dom('div').isDisabled();
+
+        expect(assert.results).toEqual([{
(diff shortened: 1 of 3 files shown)


Turbo87:
nice work @billybonks!

I've just read up on `aria-disabled` and I have a few thoughts:

- `aria-disabled` could also be applied to a parent container so we would have to traverse all parent elements until `rootElement` and check all of them for `aria-disabled`

- According to the examples at http://maxdesign.com.au/jobs/sample-accessibility/05-forms/attribute-disabled.html it seems that `disabled` and `aria-disabled` behave differently. I'm somewhat hesitant to include `aria-disabled` in the existing `isDisabled()` assertion due to that.

I'm wondering if something like `assert.dom('.foo').aria.isDisabled()` would make more sense 🤔

billybonks:
These are two very good points that i did not think about. i experimented with point 2, let me review usage in my app and point 1 then get back to you.

billybonks:
These are two very good points that i did not think about. i experimented with point 2, let me review usage in my app and point 1 then get back to you.

billybonks:
so the idea with aria-disabled is to use it on items that are not disableable by default and to implement the disable feature. so using it on a button does not make sense. aria is a fallback for things that cannot be described semantically or custom components.

So taking the example of my use case. power select it is made of generic divs, so it can't use disable attribute. in order to cirumvent this it uses aria disabled and implements the disable functionallity. so it would make sense to call it using the same api as normal disabled

So while it does work slightly differently with HTML. the places where you would need to use aria-disabled.

i do believe this will be the general use case, so putting it under an aria object would make the API a bit confusing.

Turbo87:
I guess for the sake of API simplicity I tend to agree. We should still cover the disabled parent container case though.

billybonks:
I can add that, should the message be modified to match like is disabled
because parent node was disabled ?

billybonks:
@Turbo87 i have made the basic changes, best reviewed commit by commit, last 3 commits.

one thought i had was 
```html
<div aria-disabled="true">
    <div aria-disabled="false">
       <div id="child"></div>
    </div>
</div>
```

i presume in this case child is disabled.

Turbo87:
@billybonks it looks like the tests have merge conflicts for some reason. sorry if I broken anything 😞

billybonks:
No worries I'll do a rebase in an hour or so.

lib/__tests__/is-disabled.ts
```diff
@@ -133,6 +175,6 @@ describe('assert.dom(...).isDisabled()', () => {
     expect(() => assert.dom(undefined).isDisabled()).toThrow('Unexpected Parameter: undefined');
     expect(() => assert.dom({}).isDisabled()).toThrow('Unexpected Parameter: [object Object]');
     expect(() => assert.dom(document).isDisabled()).toThrow('Unexpected Parameter: [object Document]');
-    expect(() => assert.dom(document.createElement('div')).isDisabled()).toThrow('Unexpected Element Type: [object HTMLDivElement]');
+    expect(() => assert.dom(document.createElement('div')).isDisabled()).toThrow('Generic Element Type: [object HTMLDivElement] does not use aria-disabled attribut');
```
MelSumner: May I suggest instead of "does not use" maybe try "should use" instead? It seems more consistent with [RFC 2119](https://tools.ietf.org/html/rfc2119) style of language. I know this isn't an RFC but the consistency feels nice.

billybonks: 👍 a bit late here, but thanks will add the change

lib/assertions/is-disabled.ts
```diff
@@ -13,12 +14,19 @@ export default function isDisabled(message, options: { inverted?: boolean } = {}
       element instanceof HTMLOptionElement ||
       element instanceof HTMLFieldSetElement
     )) {
-    throw new TypeError(`Unexpected Element Type: ${element.toString()}`);
+      let ariaDisabled = element.attributes['aria-disabled'];
+      if(ariaDisabled) {
+        state = ariaDisabled.value === 'true' ? true : false;
+      } else {
+        throw new TypeError(`Generic Element Type: ${element.toString()} does not use aria-disabled attribute`);
```
billybonks: can generics be disabled

billybonks: been a while

scalvert:
This PR seems like a great first step in discussing the ability to assert on ARIA attributes. I can't help but think that we need a larger, more detailed design on how this would look as *first-class* support in this library. Specifically, I'd much prefer if we supported [more ARIA specific attributes](https://www.w3.org/TR/html-aria/#aria-table) than just `aria-disabled`.

I also really like @Turbo87's idea of keying off an `aria` property, as that helps keep the assertions _targeted_ and _specific_. This is good for a number of reasons:

- helps keep the assertion logic simple and manageable
- helps you compose assertions so you're not forcing folks to assert on _all the things_ at once
- will allow for more targeted assertions that are specific (and separate to), ARIA assertions, as in some cases there's no clean overlap with an existing assertion and an aria one, as this PR proposes.

I'd like other's thoughts here.

scalvert:
@billybonks thanks for your work on this PR. Do you think we should close this in favor of https://github.com/simplabs/qunit-dom/pull/679, since the latter is a more generalized solution for asserting the presence of aria attributes?

cc/ @Turbo87

Issue #104, opened 2018-07-25
White-space is unwieldy during includesText assertion


brookjordan:
### Issue
While building HTML we regularly leave line-breaks and various inline elements, and that shouldn't interfere with testing.

The current implementation of `includesText` includes all types of whitespace and line-breaks which can make assertions very cumbersome.

This would also have the side effect of bringing `includesText` in-line with `hasText`.

### Possible solution
All white space inside the element is collapsed into a single space, similarly to how it happens in HTML.

### Possible implementation
Update this:
https://github.com/simplabs/qunit-dom/blob/632205606020d9d7e6aa3c0a83bb34502a65f9e5/lib/assertions.ts#L559-L565

to:
```js
let collapsedText = collapseWhitespace(element.textContent);
let result = collapsedText.indexOf(text) !== -1;
let actual = collapsedText;
let expected = text;

if (!message) {
  message = `Element ${this.targetDescription} has text containing "${text}"`;
}
```

And apply a similar change to: `doesNotIncludeText`

### References
- https://github.com/tradegecko/ember-semantic-test-helpers/issues/18

<sub>Edited upon finding the `collapseWhitespace` function</sub>

Turbo87:
nice find! seems like good thing to fix. it is a breaking change though, so we will need to bump the version for this.

Turbo87:
fixed by #105

PR #105, opened 2018-07-25
Collapse whitespace for `includesText` assertion

brookjordan:
Trim white-spaces from element text before comparing to expected result

see: #104


Turbo87:
can you add corresponding test for those cases?

brookjordan:
I've repeated all of the original tests but with irregular spacing.

Do you think it's right to duplicate these, or should I merge them?

Turbo87:
looks good, thanks! :)

PR #140, opened 2018-09-20
only install assertions when QUnit is defined

dwickern:
`qunit-dom` was installed for me when I upgraded from ember-cli 3.1 to 3.2. I use mocha so this caused my tests to fail.

```
qunit-dom.js:5 Uncaught ReferenceError: QUnit is not defined
    at qunit-dom.js:5
```
(followed by a bunch of cascading errors)

This change makes qunit-dom a no-op if there is no QUnit.

### DIFF
--- a/lib/qunit-dom.ts
+++ b/lib/qunit-dom.ts
@@ -6,7 +6,9 @@ declare global {
   }
 }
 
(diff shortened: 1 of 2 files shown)


Turbo87:
closing due to inactivity

PR #106, opened 2018-07-25
Add `.hasCSS()` assertion

selvagsz:
Add assertions for styles like 

```js
assert.dom('.some-element').hasCSS({
  opacity: 1,
  width: '340px',
});
```

Fixes #90

PR #168, opened 2018-10-17
Add `assert.dom(...).hasStyle()` assertion

### LINKED ISSUE #90: Add `assert.dom(...).hasCSS({ ... })` assertion
### Example

```js
assert.dom('.some-element').hasCSS({
  opacity: 1,
  width: 340,
});
```

selvagsz:
Add assertions for styles like 

```js
assert.dom('.some-element').hasStyle({
  opacity: 1,
  width: '340px',
});
```

Fixes #90 

supersedes #106

### DIFF
--- a/lib/__tests__/has-style.ts
+++ b/lib/__tests__/has-style.ts
@@ -24,6 +24,25 @@ describe('assert.dom(...).hasStyle()', () => {
     }]);
   });
 
+  test('succeeds for checking styles applied by CSS stylesheets', () => {
+    var styleNode = document.createElement('style');
+    styleNode.innerHTML = '.foo { color: blue }';
+    document.body.appendChild(styleNode);
+    assert.dom('.foo').hasStyle({
+      opacity: '1',
+      width: '200px',
+      'text-align': 'center',
+      color: 'blue',
+    });
+    expect(assert.results).toEqual([{
+      actual: { opacity: '1', width: '200px', 'text-align': 'center', color: 'blue' },
+      expected: { opacity: '1', width: '200px', 'text-align': 'center', color: 'blue' },
+      message: 'Element .foo has style \"{\"opacity\":\"1\",\"width\":\"200px\",\"text-align\":\"center\",\"color\":\"blue\"}\"',
+      result: true,
+    }]);
+    document.body.removeChild(styleNode);
+  });
+
   test('succeeds for partial style checking', () => {
     assert.dom('.foo').hasStyle({
       opacity: '1',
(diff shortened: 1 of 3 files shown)


selvagsz:
@Turbo87 this is how it renders on failure

image

selvagsz:
Just saw there is yarn docs to generate the documentation. Will use it & update the pr

selvagsz:
@Turbo87 fixed the docs & updated the tests. lmk if I'd missed anything else :)

lib/assertions.ts
```diff
@@ -430,6 +430,40 @@ export default class DOMAssertions {
...
+  hasStyle(expected: object, message?: string): void {
+    let element = this.findTargetElement();
+    if (!element) return;
+    let actual = {};
+    let result = false;
+    let cssStyleDeclaration = window.getComputedStyle(element);
+    for (let property in expected) {
+      actual[property] = cssStyleDeclaration[property];
+      if (!result) {
+        result = '' + expected[property] === actual[property];
```
Turbo87: what is this doing?

maybe something like this might be easier to read/understand:
```js
let computedStyle = window.getComputedStyle(element);

let expectedProperties = Object.keys(expected);
let result = expectedProperties.every(property => computedStyle[property] === expected[property]);

let actual = {};
expectedProperties.forEach(property => actual[property] = computedStyle[property]);
```

lib/__tests__/has-style.ts
```diff
@@ -0,0 +1,56 @@
...
+      width: '200px',
+      'text-align': 'center',
+    });
+    expect(assert.results).toEqual([{
+      actual: { opacity: '1', width: '200px', 'text-align': 'center' },
+      expected: { opacity: '1', width: '200px', 'text-align': 'center' },
+      message: 'Element .foo has style \"{\"opacity\":\"1\",\"width\":\"200px\",\"text-align\":\"center\"}\"',
+      result: true,
+    }]);
+  });
```
Turbo87: can you add another test for a partial check? right now the element has three properties, and we check exactly those properties, but we should make sure that it also works if we only check a subset of those properties.

I'm not sure if jsdom can do that but we should also test what happens if we define the `.foo` CSS class, assign properties in there and then check them with `hasStyle()`

jessica-jordan:
@Turbo87 @selvagsz Is it alright if I pick this one up? It would be sweet to see the work that has already been done land :sparkles:

gabrielcsapo:
@jessica-jordan please 🙏 this would be awesome to have!

lib/assertions.ts
selvagsz: refactored as suggested 👍

selvagsz:
Been working on product releases for a while. Sorry for the delay

lib/__tests__/has-style.ts
```diff
@@ -24,6 +24,18 @@ describe('assert.dom(...).hasStyle()', () => {
     }]);
   });
 
+  test('succeeds for partial style checking', () => {
```
jessica-jordan: 👍

Turbo87:
nice work, thanks @selvagsz!

Issue #197, opened 2018-11-16
New implementation of includesText fails when expected text contains line breaks


happycollision:
After the change made to satisfy issue #104, you can no longer pass in text containing line breaks (or probably multiple spacing) as the expected text in `includesText`.

This is not a clear limitation in the documentation, and the resulting message in Ember is confusing:

<img width="572" alt="screen shot 2018-11-15 at 8 39 31 pm" src="https://user-images.githubusercontent.com/3663628/48597191-2264ee80-e922-11e8-860d-b995aa65fae1.png">

As you can see, it appears as though the test is failing because of the diff at the beginning of the string. If you are not distracted by the red highlight in the diff, you might notice that there is actually a line break discrepancy, too.

Turbo87:
looks like this can be closed as #198 was merged and release a while ago

PR #198, opened 2018-11-16
Warn user when `includesText` assertion should expect collapsable whitespace

happycollision:
I added three tests. The first two may not be desirable specifications, since passing them might cause a major version bump again. (They test that text containing expected line breaks, etc, would pass.)

The third test is an example of a possible scenario to mitigate confusion described in #197. That would be my suggestion if passing the first two is too much of a burden. Or something like that.

### DIFF
--- a/API.md
+++ b/API.md
@@ -448,6 +448,11 @@ matching the `selector` contains the given `text`, using the
 [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)
 attribute.
 
+> Note: This assertion will collapse whitespace in `textContent` before searching.
+> If you would like to assert on a string that _should_ contain line breaks, tabs,
+> more than one space in a row, or starting/ending whitespace, use the [#hasText](#hasText)
+> selector and pass your expected text in as a RegEx pattern.
+
 **Aliases:** `containsText`, `hasTextContaining`
 
 #### Parameters
--- a/lib/__tests__/includes-text.ts
+++ b/lib/__tests__/includes-text.ts
@@ -124,6 +124,48 @@ describe('assert.dom(...).includesText()', () => {
     });
   });
 
+  describe('with HTMLElement with expected spacing', () => {
+    let element;
+
+    beforeEach(() => {
+      document.body.innerHTML = '<div>foo\n  <pre>bar\n  baz</pre></div>';
+      element = document.querySelector('div');
+    });
+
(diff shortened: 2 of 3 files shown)


Turbo87:
I'm not sure what the issue is here. `hasText()` and `includesText()` should behave similarly in general and also with regard to whitespace collapsing. I agree though, that it might make sense to print a warning somewhere if the input string contains whitespace that would be collapsed 🤔 

If your issue is that you would like to do a `has/includesText()` assertion without the automatic collapsing then maybe it would make sense to introduce a `collapseWhitespace: false` option on those assertions.

Turbo87:
> Or should it throw and error with a message in there, because bad input was supplied as the expected test?

I think we should probably just use `console.warn()` 🤔

happycollision:
Great. I think the assertions you have will work just fine. Just that helpful messaging will be sufficient.

I'll do some further tests with changes in an Ember environment, then adjust this PR.

happycollision:
After checking on this inside Ember's testing env, I still think it is possibly more useful to append the warning to the message that is pushed into the results of the assertion. I know that I do not open the inspector when I am testing unless I am looking for something that I am already expecting to see there. Things like `console.log`s that I have in my code.

To illustrate, here are two screenshots. The first shows adding a warning in the console only.
screen shot 2018-11-18 at 3 37 53 pm

You can see the message is there, but the context for it is difficult to discern. And if you have multiple test failures, you may not know which one it is referring to.

This screenshot shows when we append the warning to the message.
screen shot 2018-11-18 at 3 51 37 pm

Here, we don't need to open the console to see our message associated with the proper test.

This is achieved through adding this code to the assertion just before pushing the message:

```ts
    if (!result && text !== collapseWhitespace(text)) {
      message = message + '\n\nYour expected text contains spacing that is not preserved in this assertion. Try the `.hasText()` assertion passing in your expected text as a RegEx pattern.';
    }
```

So this code won't even run if the result is `true`.

I am happy to simply change this to `console.warn` as you suggested, though, if you still prefer that. Let me know if the messaging is what you prefer.

happycollision:
I have also updated the API docs to contain a warning to the same effect. Just lets users know that whitespace is collapsed in this assertion.

happycollision:
Should be all set, @Turbo87.

Turbo87:
thanks :)

PR #199, opened 2018-11-18
Docs: Add missing Markdown links for [ShorthandLinkStyle][]

happycollision:
There are links missing in the current documentation that are formatted
like `[HTMLElement][]`.

I originally tried to add them by just appending
the corresponding markdown to the `documentation.yml` file.
Unfortunately, when built, the link keys were all `.toLowerCase()`ed.
So I have added a very simple node script to append all necessary
links to the `API.md` file.

This script has been added to the `yarn docs` command, so
introduces no new dependencies.

### DIFF
--- a/API.md
+++ b/API.md
@@ -538,3 +538,6 @@ Assert that the `value` property of an [HTMLInputElement][] is empty.
 ```javascript
 assert.dom('input.username').hasNoValue();
 ```
+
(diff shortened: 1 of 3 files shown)


Turbo87:
This seems like a hot fix that's just hiding the actual bug. Is this a reported bug with documentation.js? If not, have we configured something wrong which makes this not work automatically?

PR #204, opened 2018-11-25
Adds support for `count` option to `isVisible` assertion

lukemelia:
- allows passing { count: N } to isVisible, with similar semantics to how exists supports a `count` option
- refactors `isNotVisible` to be implemented as `isVisible({ count: 0 })`
- introduce private `findElements` method to be used similarly to `this.findElement`
- refactor to use `findElements` in `exists` as well as in `isVisible`


lukemelia:
BTW, the practical driver for this was working today with a virtual list component where some DOM nodes are kept around with `display: none` for future reuse, but I have wished for this option in the past in other situations as well.

Issue #206, opened 2018-11-30
add a matchesSelector() and doesNotMatchSelector() assertion


timo-lindemann-victors:
Would it be a good idea to have something like

```js
assert.dom(selector).matchesSelector(anotherSelector, message)
```

to be able to check if a given CSS selector returns the exact same (set of?) element(s)?

I would probably save some boilerplate if we had that, and CSS selectors have come to express a lot more things than just CSS classes and attributes (sibling detection, last-child, ...)

Just checking the waters here.

Turbo87:
sound like a good idea 👍 

two more thoughts:

1. the snippet above uses `matches()` but as the issue title suggests we should use `matchesSelector()` to be a little more specific

2. if we have an `<div id="foo" class="red">` and other uses of `red` on the page, what would `assert.dom('#foo').matchesSelector('.red')` result in?

timo-lindemann-victors:
@Turbo87 glad you like it.

 Thanks for pointing out my mistake; I have corrected the name of the assertion in my opening question.

To your question: The assertion should use [Element.matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) so the answer is most likely a passing assertion for `matchesSelector ` and a failure for `doesNotMatchSelector`.

So, if the `anotherSelector` returns a set of elements containing the element of the first selector, the assertion passes.

I have not decided what to do if `selector` describes multiple elements. I think the most sensible approach is to check that the result set of `selector` is s subset of `anotherSelector`.

Thoughts?

lindem:
I might play around with it on sunday.

timo-lindemann-victors:
Fixed by #208.

PR #208, opened 2018-12-01
Add `matchesSelector()` assertions

lindem:
working at #206 please comment


lib/assertions/matches-selector.ts
```diff
@@ -0,0 +1,15 @@
+export default function matchesSelector(
+  targetSelector: string | Element,
+  compareSelector: string
+): Array<number> {
+  let targetElements: Array<Element> =
+    typeof targetSelector === 'string'
+      ? Array.from(document.querySelectorAll(targetSelector))
+      : [targetSelector];
```
Turbo87: this code should probably reuse the `findElements()` method of the `Assertions` class. I also think we should just pass an `Element[]` array to this function, and let the surrounding code worry about dealing with the target selectors.

lib/assertions/matches-selector.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  let failures = targetElements.reduce(function(acc, element) {
+    return element.matches(compareSelector) ? acc : acc + 1;
+  }, 0);
```
Turbo87:
```js
let failures = targetElement.filter(it => it.matches(compareSelector)).length > 0;
```
👆 seems a little easier to read to me

lib/__tests__/matches-selector.ts
```diff
@@ -0,0 +1,123 @@
...
+      expect(assert.results).toEqual([{
+        actual: 2,
+        expected: 3,
+        message: '1 out of 3 elements selected by p did not also match the selector p + p.',
+        result: false
+      }]);
+    });
+  })
+
+  describe('doesNotMatchSelector: success scenarios', () => {
```
Turbo87: `doesNotMatchSelector` should have it's own dedicated testing file and not be nested in the `assert.dom(...).matchesSelector()` test suite

lib/__tests__/matches-selector.ts
```diff
@@ -0,0 +1,123 @@
...
+      assert.dom('p').matchesSelector('div>p');
+
+      expect(assert.results).toEqual([{
+        actual: 3,
+        expected: 3,
+        message: '3 elements, selected by p, also match the selector div>p.',
+        result: true
+      }]);
+    });
+  });
```
Turbo87: this is missing test cases where an `Element` is passed directly to `assert.dom(...)`. have a look at the other test suites for examples.

lindem: So I did accomodate instances of Element passed to the assertion and looked at how hasClass does it.

lib/__tests__/matches-selector.ts
lindem: I placed these tests into another file and also reduced redundancy in wording.

lib/assertions/matches-selector.ts
lindem: agreed, but to count failures, we filter the negative results, making it `!it.matches(...)`.

lib/assertions/matches-selector.ts
lindem: I applied the suggested changes.

lindem:
Done with it again (for now). I am unsure if the new redundancy in the assertions.ts file is warranted (doesNotMatchSelector and matchesSelector have a lot in common there), but it works and is easy to understand I hope.

Still happy to continue working on it. Thanks for patience.

lib/__tests__/does-not-match-selector.ts
```diff
@@ -0,0 +1,84 @@
...
+    document.body.innerHTML = '<div><p class="first"></p><p></p><p class="last"></p></div>';
+  });
+
+  describe('success scenarios', () => {
+    test('one element (selector passed) not matching compareSelector', () => {
+      assert.dom('p.first').doesNotMatchSelector('p + p');
+
+      expect(assert.results).toEqual([{
+        actual: 0,
+        expected: 0,
```
Turbo87: we should use better `actual/expected` values. have a look at some of the other assertions, we often reuse the assertion message, and generate another assertion message for `expected`.

lindem: I have been coming down with severe health problems during the weekend. I will get to it once I am well enough to code again. Sorry for the delay.

Turbo87: no worries, gute besserung 😉

lindem:
@Turbo87 I hope this is more what you wanted. I am seemingly unable to resolve the merge conflict in this PR myself (it is not clear that I *can* do something about it from the page the button leads to, but then, I cannot write to your repository).

Ah, and happy new year.

lib/__tests__/does-not-match-selector.ts
lindem: thanks very much.

PR #213, opened 2018-12-07
Replace `Array.from()` for IE11 compatibility

Turbo87:
Resolves https://github.com/simplabs/qunit-dom/issues/212

/cc @jelhan

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -836,7 +836,7 @@ export default class DOMAssertions {
       return [];
 
     } else if (typeof this.target === 'string') {
-      return Array.from(this.rootElement.querySelectorAll(this.target));
+      return [].slice.call(this.rootElement.querySelectorAll(this.target));
 
     } else {
       throw new TypeError(`Unexpected Parameter: ${this.target}`)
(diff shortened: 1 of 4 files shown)


lib/assertions.ts
```diff
@@ -836,7 +836,7 @@ export default class DOMAssertions {
       return [];
 
     } else if (typeof this.target === 'string') {
-      return Array.from(this.rootElement.querySelectorAll(this.target));
+      return [].slice.call(this.rootElement.querySelectorAll(this.target));
```
jelhan: Doesn't make this code more readable IMO. Maybe adding an `arrayFrom()` utility function to hide that complexity?

mansona: Personally, I think a regression in readability is worth it if it fixes the experience for IE11. And I also wouldn't like to add more code than was necessary to get the job done, adding a utility function would make this PR much larger 🙈

simonihmig: Maybe `Array.prototype.slice.call()` would be slightly better performing, as it would not create a useless empty array that would need to be garbage collected? But probably neglectable...

A quick bug fix release would be appreciated though! 😁

PR #228, opened 2018-12-26
Fix visibility check

kratiahuja:
Ideally an element is not visible when both its offsetHeight and offsetWidth is 0. You could have elements that have a width but not height.
The [docs](https://github.com/simplabs/qunit-dom/blob/master/API.md#isvisible) say `and` but we check for `or`.

cc: @Turbo87

### DIFF
--- a/lib/helpers/visible.ts
+++ b/lib/helpers/visible.ts
@@ -3,13 +3,13 @@
 
 export default function visible(el) {
   if (el === null) return false;
-  if (el.offsetWidth === 0 || el.offsetHeight === 0) return false;
+  if (el.offsetWidth === 0 && el.offsetHeight === 0) return false;
 
   let clientRects = el.getClientRects();
   if (clientRects.length === 0) return false;
   for (let i = 0; i < clientRects.length; i++) {
     let rect = clientRects[i];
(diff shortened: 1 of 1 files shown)


Turbo87:
I think the docs might be wrong on this one. If either width or height are zero then the size is zero too and the element will be invisible. or am I missing a special case?

kratiahuja:
There is a special case (atleast what I can see) is where the height is 0 but width isn't 0 and the element is visible. The other issue here is also that this started breaking after 0.6.3 with this [diff](https://github.com/simplabs/qunit-dom/commit/87f44839e1298c1ed4374f23b90e70dc86e5aa4c). The earlier diff allowed this condition to be marked as visible v/s now it doesn't.

Turbo87:
Hmm, how can something without height be visually visible?

kratiahuja:
If the element is displaced from its original place but is still "visible", then offsetHeight seems to be 0 in that case (atleast from what I can see from my local debugging).

Turbo87:
do you have an example anywhere? I'm not sure how to reproduce that 🤔

Turbo87:
closing due to inactivity

PR #232, opened 2019-01-03
Implement a `hasPseudoElementStyle` assertion

joankaradimov:
Resolves https://github.com/simplabs/qunit-dom/issues/229

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -492,15 +492,39 @@ export default class DOMAssertions {
    * @see {@link #hasClass}
    */
   hasStyle(expected: object, message?: string): void {
+    this.hasPseudoElementStyle(null, expected, message);
+  }
+
+
+  hasPseudoElementStyle(selector: string, expected: object, message?: string): void;
+
+  /**
+   * Assert that the pseudo element for `selector` of the [HTMLElement][] has the `expected` style declarations using
+   * [`window.getComputedStyle`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle).
+   *
+   * @name hasPseudoElementStyle
+   * @param {string} selector
+   * @param {object} expected
+   * @param {string?} message
+   *
+   * @example
+   * assert.dom('.progress-bar').hasPseudoElementStyle(':after', {
+   *   content: '";"',
+   * });
+   *
+   * @see {@link #hasClass}
+   */
+  hasPseudoElementStyle(selector: string | null, expected: object, message?: string): void {
     let element = this.findTargetElement();
     if (!element) return;
-    let computedStyle = window.getComputedStyle(element);
+    let computedStyle = window.getComputedStyle(element, selector);
     let expectedProperties = Object.keys(expected);
     let result = expectedProperties.every(property => computedStyle[property] === expected[property]);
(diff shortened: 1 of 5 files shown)


lib/assertions.ts
```diff
@@ -492,15 +492,41 @@ export default class DOMAssertions {
...
+   * @param {string?} message
+   *
+   * @example
+   * assert.dom('.progress-bar').hasPseudoElementStyle(':after', {
+   *   content: '";"',
+   * });
+   *
+   * @see {@link #hasClass}
+   */
+  hasPseudoElementStyle(selector: string, expected: object, message?: string): void {
```
Turbo87: I **think** this would have to be `selector: string | null`, and then we'd need another variant above that accepts `selector: string` as the public interface to that method. If I remember correctly the actual implementation is always hidden if you have specializations like the `selector: null` above 🤔

joankaradimov: Will do.

joankaradimov:
> nice work @joankaradimov!

Thanks!

> just one minor thing left to do :)

I've committed the suggested change.

Just let me know if everything else looks good so that I squash the commits.

lib/assertions.ts
```diff
@@ -492,15 +492,42 @@ export default class DOMAssertions {
...
+   * @param {object} expected
+   * @param {string?} message
+   *
+   * @example
+   * assert.dom('.progress-bar').hasPseudoElementStyle(':after', {
+   *   content: '";"',
+   * });
+   *
+   * @see {@link #hasClass}
+   */
```
Turbo87: can you check that `yarn docs` reads the comment correctly? the `API.md` change should then also be committed, I guess

joankaradimov: It does not work correctly.

joankaradimov:
It turns out the initial overload signature for `hasPseudoElementStyle` was the correct one. The one in a858a7386012f7aa7779fc47cc84f5ea3aa997f9 does not generate any documentation. I've reverted it.

Also there seemed to be some code which did not have a documentation generated for it. I've committed that separately.

Finally - there are some DOS new lines mixed with the Unix new lines in the examples in API.md. This appears to be the case for the existing examples too. There are no DOS new lines in the docstrings. Should I log a bug for that?

Turbo87:
> there are some DOS new lines mixed with the Unix new lines in the examples in API.md. This appears to be the case for the existing examples too. There are no DOS new lines in the docstrings. Should I log a bug for that?

Interesting. That sounds like a bug (or maybe even feature 🤔) in `documentation` to me

Turbo87:
@joankaradimov the problem with `documentation` is that is does not pick up the docs from these method specializations. I've moved the docstring on the method implementation, while keeping the `selector: string` specialization and now it seems to work as intended :)

joankaradimov:
> I've moved the docstring on the method implementation

Ah, ok. I was thinking of something different.

> there seemed to be some code which did not have a documentation generated for it.

I see you made this private and removed the commit.

I'm going to squash the rest of it and force push.

Turbo87:
thanks again 🙏

joankaradimov:
Thank you for the help and the quick responses. And also for quickly releasing this!

PR #246, opened 2019-01-14
Make `findTargets` support a `this.target` of type `Element`

### LINKED ISSUE #245: assert.dom(Element).exists() throws even if it exists
Even though officially the `dom` method accepts the `Element` type, when using it that way the `exists` method ends up throwing an error, even if the element exists. At root, this is because `exists` calls `findElements`, which is not currently set up to check for `this.target` being anything other than a string.

I have a PR inbound shortly to add support to `findElements` for having Element targets, but wanted to track it here also.

dwilhelmi:
Resolves #245

### DIFF
--- a/lib/__tests__/is-visible.ts
+++ b/lib/__tests__/is-visible.ts
@@ -17,7 +17,7 @@ describe('assert.dom(...).isVisible()', () => {
     assert = new TestAssertions();
   });
 
-  describe('selector only', () => {
+  describe('with selector', () => {
     test('fails if element is missing', () => {
       document.body.innerHTML = '<h1 class="baz">foo</h1>bar';
 
@@ -45,6 +45,25 @@ describe('assert.dom(...).isVisible()', () => {
           actual: 'Element h1 is not visible',
           expected: 'Element h1 is visible',
           message: 'foo',
+          result: false,
+        },
+      ]);
+    });
+  });
+
+  describe('with Element', () => {
+    test('fails for missing element', () => {
+      document.body.innerHTML = '<h1 class="baz">foo</h1>bar';
+
+      const headingElement = document.querySelector('h2');
+      console.log(headingElement);
+      assert.dom(headingElement).isVisible();
+
+      expect(assert.results).toEqual([
+        {
+          actual: 'Element <not found> is not visible',
+          expected: 'Element <not found> is visible',
+          message: 'Element <not found> is visible',
           result: false,
         },
       ]);
--- a/lib/__tests__/is-visible.ts
+++ b/lib/__tests__/is-visible.ts
@@ -32,6 +32,21 @@ describe('assert.dom(...).isVisible()', () => {
         },
       ]);
     });
+
+    test('fails if element is missing', () => {
+      document.body.innerHTML = '<h1 class="baz">foo</h1>bar';
+
+      assert.dom('h2').isVisible();
+
+      expect(assert.results).toEqual([
+        {
+          actual: 'Element h2 is not visible',
+          expected: 'Element h2 is visible',
+          message: 'Element h2 is visible',
+          result: false,
+        },
+      ]);
+    });
   });
 
   describe('custom messages', () => {
@@ -55,9 +70,7 @@ describe('assert.dom(...).isVisible()', () => {
     test('fails for missing element', () => {
       document.body.innerHTML = '<h1 class="baz">foo</h1>bar';
 
-      const headingElement = document.querySelector('h2');
-      console.log(headingElement);
-      assert.dom(headingElement).isVisible();
+      assert.dom(null).isVisible();
 
       expect(assert.results).toEqual([
         {
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -1196,16 +1196,18 @@ export default class DOMAssertions {
   }
 
   /**
-   * Finds a collection of HTMLElement instances from target using querySelectorAll
+   * Finds a collection of Element instances from target using querySelectorAll
    * @private
-   * @returns (HTMLElement[]) an array of HTMLElement instances
+   * @returns (Element[]) an array of Element instances
    * @throws TypeError will be thrown if target is an unrecognized type
    */
-  private findElements(): HTMLElement[] {
+  private findElements(): Element[] {
     if (this.target === null) {
       return [];
     } else if (typeof this.target === 'string') {
       return toArray(this.rootElement.querySelectorAll(this.target));
+    } else if (this.target instanceof Element) {
+      return [this.target];
     } else {
       throw new TypeError(`Unexpected Parameter: ${this.target}`);
     }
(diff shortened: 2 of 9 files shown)


lib/assertions/exists.ts
```diff
@@ -3,15 +3,21 @@ export default function exists(options, message) {
     message = options;
     options = undefined;
   }
+  let label;
+  if (typeof this.target === 'string') {
+    label = this.target;
+  } else if (this.target instanceof Element) {
+    label = (this.target as Element).tagName;
+  }
```
Turbo87: you might want to reuse the existing `targetDescription` property for this. but we should make sure to write tests for all the valid inputs (if they don't all exist yet) and check if the assertion messages are correct.

lib/assertions.ts
```diff
@@ -965,16 +965,18 @@ export default class DOMAssertions {
...
    * @throws TypeError will be thrown if target is an unrecognized type
    */
-  private findElements(): HTMLElement[] {
+  private findElements(): Element[] {
     if (this.target === null) {
       return [];
     } else if (typeof this.target === 'string') {
       return toArray(this.rootElement.querySelectorAll(this.target));
+    } else if (this.target instanceof Element) {
+      return [this.target];
```
Turbo87: `findElements()` is also used by the `isVisible()` assertion, I assume we would have to add the `format()` label change there too

scalvert:
Any chance we can dust off this PR? It looks like there may be some straightforward changes to make.

lib/assertions/exists.ts
scalvert: I've moved some of this logic into `targetDescription`. Feel free to critique there!

lib/assertions.ts
scalvert: I made sure this worked for both `exists` and `visible`.

scalvert:
@Turbo87 in light of this PR, thoughts on changing all references to `Element` instead of `HTMLElement`? It feels like that base class is more accurate to what the code is using.

scalvert:
Once CI passes this is GTG.

scalvert:
I'll follow up this PR with one that renames all references to `HTMLElement` to `Element`, so that we're correctly reflecting the type in the comments.

scalvert:
@Turbo87 do you have an concerns with me merging this. Do you need to get 👀 on it? I feel as if it's mostly additive.

alexhancock:
New to the conversation, but I would like to see this merged as well

lib/__tests__/is-visible.ts
```diff
@@ -51,6 +51,25 @@ describe('assert.dom(...).isVisible()', () => {
     });
   });
 
+  describe('with Element', () => {
+    test('fails for missing element', () => {
+      document.body.innerHTML = '<h1 class="baz">foo</h1>bar';
+
+      const headingElement = document.querySelector('h2');
+      console.log(headingElement);
```
Turbo87: 🤔

lib/__tests__/is-visible.ts
```diff
@@ -51,6 +51,25 @@ describe('assert.dom(...).isVisible()', () => {
...
+      assert.dom(headingElement).isVisible();
```
Turbo87: I guess we could just pass `undefined` directly here instead of doing the `querySelector()` dance above

lib/assertions.ts
```diff
@@ -1063,6 +1065,12 @@ export default class DOMAssertions {
...
    */
   private get targetDescription() {
-    return elementToString(this.target);
+    if (typeof this.target === 'string') {
+      return this.target;
+    } else if (this.target instanceof Element) {
+      return elementToString(this.target);
+    } else {
+      return '<not found>';
+    }
```
Turbo87: why did you adjust `targetDescription()` instead of adding this to `elementToString()`?

tests/acceptance/qunit-dom-test.js
```diff
@@ -34,6 +34,9 @@ module('Acceptance | qunit-dom', function(hooks) {
     assert.dom('#hidden-input').isNotVisible();
     assert.dom('p').isVisible({ count: 2 });
 
+    let titleElement = document.querySelector('#title');
+    assert.dom(titleElement).isVisible();
```
Turbo87: I don't think we need this test here. the tests here are essentially just smoke tests to see if the library is loaded correctly. there is no need to test every single feature here.

lib/__tests__/is-visible.ts
scalvert: Oops.

lib/__tests__/is-visible.ts
scalvert: We totally could. I guess this was existing?

scalvert: I think passing `null` is more appropriate, as that's what's returned if an element isn't found.

tests/acceptance/qunit-dom-test.js
scalvert: Fair. I'll remove, since we cover `isVisible`/`isNotVisible` above adequately.

lib/assertions.ts
scalvert: Good question. I can't remember my thought process. It _does_ feel more appropriate to add to `elementToString()`. Let me look at it again and see what I was thinking.

scalvert: Here comes the long winded reply:

This was added to address a particular case that affects both `is-visible` _and_ `exists`. It was explicitly related to fixing `is-visible`'s tests when an element is not found via `document.querySelector`, and therefore returns `null`. This is not present in the case where a `string` selector is passed. Specifically, this test:

https://github.com/simplabs/qunit-dom/blob/29cade3b969949f1d535774b79225653263c6bc4/lib/__tests__/is-visible.ts#L58-L60

that you called out in a comment above _does_ have value. That's because we don't test for a case where you pass an expected element as the target, but that element isn't found in the DOM. 

The `exists` tests only cover the following case:

https://github.com/simplabs/qunit-dom/blob/29cade3b969949f1d535774b79225653263c6bc4/lib/__tests__/exists.ts#L31

so the `is-visible` tests are slightly more expansive.

The reason this affects `exists` and `is-visible` is that both utilize the following pattern in their assertions:

https://github.com/simplabs/qunit-dom/blob/29cade3b969949f1d535774b79225653263c6bc4/lib/assertions/exists.ts#L7

where they try to find all elements that match the target in order to get a count. When the target is `null`, this returns an empty array via `this.findElements`:

https://github.com/simplabs/qunit-dom/blob/29cade3b969949f1d535774b79225653263c6bc4/lib/assertions.ts#L1053-L1054

So, when the code continues within the respective assertions for `exists` and `is-visible`, we try to utilize `this.targetDescription` to format the description. The `targetDescription` getter expects either a `string` or an `Element`, but in the above case the target is neither; it's `null`. Since it's not an `Element`, it made sense to add this as a fallback here, inside `targetDescription`, so that we can cover any case where `this.targetDescription` is used on a target that's neither a `string` or `Element`. I feel like this is a good fallback. 

The only thing I'd add to this PR at this point is a test to ensure that `exists` does the correct thing when `null` is passed.

PR #365, opened 2019-05-16
Specify options.count for exists and isVisible

steventsao:
Be explicit on the properties available in options.


Turbo87:
@steventsao that file is automatically generated. without a change to the sourcecode your change would be overwritten the next time the file is regenerated.

steventsao:
Thanks. Submitted another PR with fix https://github.com/simplabs/qunit-dom/pull/368/

PR #451, opened 2019-08-20
Adds regex support for CSS class assertions

### LINKED ISSUE #450: Regex support for hasClass / doesNotHaveClass
I recently needed to use a regex to match against a CSS class, which `hasClass` and `doesNotHaveClass` don't currently support.

Would this feature be considered, if I opened a PR?

The context is:

- I'm using namespaced CSS modules, which prepend a unique string to CSS classes, like `d66d3pf_my-component_is-great`
- I don't know this `d66d3pf` prefix within my tests
- I would like to check that the CSS class is present / absent, while ignoring the prefix

Here's what I'm doing right now, within an EmberJS component rendering test:

```js
const cssClasses = this.element.classList.value
assert.ok(cssClasses.match(/my-component_is-great/))
```

timiyay:
This PR adds RegExp support to the `hasClass` and `doesNotHaveClass` assertions, as suggested in https://github.com/simplabs/qunit-dom/issues/450. 

Both functions now accept `string|RegExp`, in the same vein as `hasText`.

Each CSS class present on the target element will be matched against the RegExp, and:

- for the positive `hasClass` assertion to return `true`, at least one CSS class must match; and
- for the negative `doesNotHaveClass` assertion to return `true`, none of the CSS classes may match

Resolves #450

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -481,35 +481,55 @@ export default class DOMAssertions {
    * Assert that the {@link HTMLElement} does not have the `expected` CSS class using
    * [`classList`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList).
    *
+   * `expected` can also be a regular expression, and the assertion will return
+   * true if none of the element's CSS classes match.
+   *
    * **Aliases:** `hasNoClass`, `lacksClass`
    *
-   * @param {string} expected
+   * @param {string|RegExp} expected
    * @param {string?} message
    *
    * @example
    * assert.dom('input[type="password"]').doesNotHaveClass('username-input');
    *
+   * @example
+   * assert.dom('input[type="password"]').doesNotHaveClass(/username-.*-input/);
+   *
    * @see {@link #hasClass}
    */
-  doesNotHaveClass(expected: string, message?: string): void {
+  doesNotHaveClass(expected: string | RegExp, message?: string): void {
     let element = this.findTargetElement();
     if (!element) return;
 
-    let result = !element.classList.contains(expected);
     let actual = element.classList.toString();
 
-    if (!message) {
-      message = `Element ${this.targetDescription} does not have CSS class "${expected}"`;
-    }
+    if (expected instanceof RegExp) {
+      let classNames = Array.prototype.slice.call(element.classList)
+      let result = classNames.every((className: string): boolean => {
+        return !expected.test(className);
+      })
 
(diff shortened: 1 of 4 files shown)


lib/assertions.ts
```diff
@@ -435,61 +435,101 @@ export default class DOMAssertions {
...
     if (!element) return;
 
     let actual = element.classList.toString();
-    let result = element.classList.contains(expected);
 
-    if (!message) {
-      message = `Element ${this.targetDescription} has CSS class "${expected}"`;
-    }
+    if (expected instanceof RegExp) {
+      let classNames = Array.prototype.slice.call(element.classList)
```
timiyay: I originally used `Array.from` here to convert the `DOMTokenList` into an array, but reverted to `Array.prototype.slice.call` when I saw it elsewhere in the codebase.

Any preference?

lib/assertions.ts
```diff
@@ -435,61 +435,101 @@ export default class DOMAssertions {
...
+      let result = classNames.some((className: string): boolean => {
```
timiyay: I'm new to Typescript - is annotating this anonymous function overkill, or is it good practice to annotate everywhere?

Turbo87: > is annotating this anonymous function overkill

as long as TypeScript knows that `classNames` is an Array then it already knows about the signature of the `.some()` method. tl;dr in most cases it is not necessary.

lib/assertions.ts
```diff
@@ -435,61 +435,101 @@ export default class DOMAssertions {
...
     if (!element) return;
 
     let actual = element.classList.toString();
-    let result = element.classList.contains(expected);
 
-    if (!message) {
-      message = `Element ${this.targetDescription} has CSS class "${expected}"`;
-    }
+    if (expected instanceof RegExp) {
+      let classNames = Array.prototype.slice.call(element.classList);
```
Turbo87: `Array.from()` is not available on Internet Explorer yet (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from), which is why we need to use this workaround 🤷‍♂

timiyay:
Thanks for the review @Turbo87 🙏

timiyay:
I've opened this PR ready to merge on an Aussie Monday morning, and realised I don't have merge permissions 😆 

@Turbo87 Is the PR mergeable as-is, or can I help with any extra work to get it over the line?

Turbo87:
whoops, sorry, forgot about this MR 🙈

PR #460, opened 2019-08-28
Use collapseWhitespace when using Regex

rwwagner90:
This seems like an inconsistency, and threw me off trying to test with a regex. This should collapse space for both regex and string now.

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -584,8 +584,8 @@ export default class DOMAssertions {
     if (!element) return;
 
     if (expected instanceof RegExp) {
-      let result = expected.test(element.textContent);
-      let actual = element.textContent;
+      let actual = collapseWhitespace(element.textContent);
+      let result = expected.test(actual);
 
       if (!message) {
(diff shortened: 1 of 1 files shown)


Turbo87:
I'm not sure I agree. Using a regex it is somewhat easy right now to account multiple whitespace, but if we collapsed it, there would be no way to disable the collapsing when you actually do want to test for multiple whitespace 🤔

rwwagner90:
@Turbo87 we should update the docs then. They indicate whitespace is collapsed.

Turbo87:
true, yeah, that statement should only be made for the string-mode

scalvert:
I believe this is addressed in the comment for `hasText`:

https://github.com/simplabs/qunit-dom/blob/2dfb164ccc0c6db49719eba6a0e33f209f4ecf01/lib/assertions.ts#L601-L603

I'm going to close this, but please feel free to reopen if you disagree/want to continue the conversation.

rwwagner90:
I think it would be nice to collapse all the time, but as long as the docs are updated to say it does not when using RegEx, fine with me 👍

Issue #493, opened 2019-10-01
Add hasTag() assertion


jelhan:
If I didn't missed something, qunit-dom does not provide yet any assertion for the `tagName` of an element. Something like this:

```js
assert.equal(document.querySelector('[data-test-foo']).tagName, 'P', 'is a paragraph');
```

I'm not opinionated on the concrete naming but something like this would be a nice addition to the API:

```js
assert.dom('[data-test-foo']).hasTag('p');
```

As shown in the example I would recommend that the check is done case-insensitive.

Turbo87:
indeed, that might be a good addition. especially the case insensitivity would be good. are you interested in working on a PR for this?

jelhan:
> are you interested in working on a PR for this?

I'm always interested but having the feeling that I'm currently trying to work on to many repos at the same time. Need to focus to actually finish some stuff. So I'm not sad if someone else is picking that one up. :smile:

scalvert:
I gotchu, @jelhan.

esantelises:
@Turbo87 do we have a date when this feature going be release ?

Turbo87:
@esantelises sorry for the delay. https://github.com/simplabs/qunit-dom/releases/tag/v0.9.1 is out now.

esantelises:
thank you @Turbo87

PR #496, opened 2019-10-03
Adding hasTagName assertion

scalvert:
Implements #493.


scalvert:
I believe I need to add an entry in the ember tests.

lib/assertions.ts
```diff
@@ -955,6 +956,57 @@ export default class DOMAssertions {
...
+
+    if (typeof tagName !== 'string') {
+      throw new TypeError(`You must pass a string to "hasTag". You passed ${tagName}.`);
+    }
+
+    if (!TAG_NAMES.includes(tagName.toLowerCase())) {
+      throw new Error(
+        `The tagName '${tagName}' is not a valid HTML tag. You must provide a valid HTML tag`
+      );
+    }
```
Turbo87: this seems a bit too restrictive to me. at some point new tag names might be introduced or people will use webcomponents that use custom tag names.

lib/assertions.ts
```diff
@@ -955,6 +956,57 @@ export default class DOMAssertions {
...
+
+    actual = element.tagName.toLowerCase();
+    expected = tagName;
```
Turbo87: we should probably apply `toLowerCase()` to the passed in `tagName` too 🤔

lib/assertions.ts
scalvert: Ya this was a toss up, and to be fair I’m happy to remove it. If the caller passes an invalid HTML tag the test will fail anyway, so maybe that’s enough?

lib/assertions.ts
scalvert: Agree. I'll clean it up a bit.

lib/assertions.ts
```diff
@@ -955,6 +955,51 @@ export default class DOMAssertions {
...
+   * @param {string?} message
+   *
+   * @example
+   * // <h1 id="title">
+   * //   Title
+   * // </h1>
+   *
+   * assert.dom('#title').hasTag('h1');
+   */
+  hasTag(tagName: string, message?: string) {
```
Turbo87: I think I would prefer `hasTagName()` 🤔 

thoughts?

scalvert: Ha, I quite literally thought the same thing. I'll make the change now.

scalvert:
@Turbo87 I forget how to make the ember tests work. I added an assertion, but `hasTagName` doesn't appear to be available in the version of `qunit-dom` the tests are using. I remember having this issue before, but can't remember how I resolved :)

scalvert:
I believe I figured it out. Added ember test.

scalvert:
I think this PR should be closely followed by `doesNotHaveTagName`

tests/acceptance/qunit-dom-test.js
```diff
@@ -18,6 +18,7 @@ module('Acceptance | qunit-dom', function(hooks) {
     assert.dom('#title').exists();
     assert.dom('#subtitle').doesNotExist();
     assert.dom('#title').hasText('Welcome to Ember');
+    assert.dom('#title').hasTagName('h2');
```
Turbo87: FWIW we don't have to add all the assertions here. this is just meant as a smoke test that the qunit-dom integration works correctly. the assertions itself are tested enough by the Jest test suite.

Issue #418, opened 2019-07-11
isVisible triggers on opacity 0


Alex-Aralis:
The current isVisible considers elements with `opacity: 0` to be visible.

This seems wrong.

Turbo87:
see https://github.com/simplabs/qunit-dom/blob/master/API.md#isvisible

steveszc:
The current implementation of `isVisible` is based on jquery's [`:visible` pseudo-selector](https://github.com/jquery/jquery/blob/4a2bcc27f9c3ee24b3effac0fbe1285d1ee23cc5/src/css/hiddenVisibleSelectors.js#L11-L13) which states the following:

```
Elements are considered visible if they consume space in the document. Visible elements have a width or height that is greater than zero.

Elements with visibility: hidden or opacity: 0 are considered visible, since they still consume space in the layout.
```
[source](https://api.jquery.com/visible-selector/)

I can see definite utility in altering `isVisible` (or introducing a new assertion) to also check for `opacity: 0` and `visibility: hidden` (maybe other properties I'm not thinking of).

Turbo87:
the problem is that there are soooo many ways to make an element invisible that it's not really viable to support all of them. it's correct that the current implementation is based on the jQuery implementation, but apparently that is not quite sufficient. I'm open to adjusting the logic, but that would most likely have to be a breaking change then.

scalvert:
We could also consider having options to `isVisible`, whereby you'd specify that you want to _also_ assert on additional properties (`opacity: 0`, `visibility: hidden`). Or, we could have a completely separate assertion that is for CSS visibility:

`isCSSVisible` - passes if `opacity` !== 0 and `visibility` !== 'hidden'
`isNotCSSVisible` - passes if `opacity` === 0 or `visibility === 'hidden'

simonihmig:
Just stumbled upon this. I was also expected `opacity` to be taken into account.

> Elements with visibility: hidden or opacity: 0 are considered visible, since they still consume space in the layout.

Tbh, that definition of visibility totally contradicts my "common sense" understanding of what visible is! If a button consumes space, but I cannot see it nor read its text, I would certainly not consider it visible! 😆
But let's not bike shed about this...

I did however remember that Selenium/Webdriver based solutions handle visibility in a IMHO smarter way. Quick google found this: https://github.com/thefrontside/element-is-visible. Maybe we can defer the decision what visible means to this (quite elaborate) util, which seems to mimic the Selenium semantics!?

[...]

PR #532, opened 2019-11-11
WIP - Adding isCSSVisible assertion

scalvert:
First pass at resolving #418. 

I'm open to having a different approach.

TODO:
- [ ] Generate docs
- [ ] Add optional options object with count (TBD)


scalvert:
@Turbo87 interested in your thoughts here. If we don't feel this has legs I'd rather just close it.

Turbo87:
> @Turbo87 interested in your thoughts here

TBH I think the best way forward might be to remove the `isVisible()` assertion completely. There are just too many ways that this could lead to false positives or negatives and I think adding a second `isVisible()` assertion variant will likely make the situation worse.

PR #542, opened 2019-11-14
Fix embroider compatibility

ef4:
This package defines an empty module for the convenience of typescript users, so they can say `import "qunit-dom"` to get the types loaded. It's not really needed for any actual code, because the code gets applied directly to QUnit's own objects.

But because this package declares itself an ember-addon, its own module namespace is expected to be defined by real modules in `treeForAddon` or `treeForAddonTestSupport`. Not a dynamic `define` in `vendor`.

`define()` is most commonly used for things that aren't in an addon's own module namespace, and that continues to work under embroider. But this more unusual case -- an addon `define()`ing itself -- is not supported.

Without this change, attempting to `import "qunit-dom"` fails under embroider.


vendor/dummy-module.js
```diff
@@ -0,0 +1 @@
+export default {};
```
Turbo87: this project does not have a dependency on `ember-cli-babel` because I wanted to keep the dependency tree for non-Ember users small. will this still be transpiled correctly without the `ember-cli-babel` dependency?

Turbo87: looking at the CI failure, I guess the answer is: nope

ef4: Oops. I forgot to test this locally *without* embroider.

I think I see a supported path, trying something else.

ef4:
I changed this so the dummy module is expressed as named AMD. This requires me to be more nuanced about what embroider supports, since I just said above that an addon can't `define()` its own modules, and it looks like my PR is doing that.

To be more precise, an addon can't `define()` its own modules *from other arbitrary files* (like `vendor/define-dummy-module.js`). But we do support the case where we're handed a `treeForAddon` or `treeForAddonTestSupport` which happens to have been precompiled to named AMD. This case works because the files are all still in their correctly resolvable locations, and the names that are being `defined()` precisely match the real filesystem layout.

vendor/dummy-module.js
```diff
@@ -0,0 +1 @@
+define('qunit-dom', function(){ return {} });
```
Turbo87: this is "missing" the `[]` argument from the previous implementation. is that optional?

ef4: It is, but it does seem silly for me to change this, now that I've gone back to using the same pattern as the old file. I will push another commit so they match exactly.

PR #587, opened 2019-12-27
TypeScript: Turn on `noImplicityAny` and fix a few type signatures

LucasHill:
I was debugging some type issues I was having and decided to turn on 'noImplicityAny' (considered a best practice in TS). This should tighten up the types to be a lot more useful. I tried to make minimal actual code changes and just do types in this pass.

### DIFF
--- a/lib/__tests__/does-not-exist.ts
+++ b/lib/__tests__/does-not-exist.ts
@@ -3,7 +3,7 @@
 import TestAssertions from '../helpers/test-assertions';
 
 describe('assert.dom(...).doesNotExist()', () => {
-  let assert;
+  let assert: TestAssertions;
 
   beforeEach(() => {
     assert = new TestAssertions();
@@ -76,10 +76,14 @@ describe('assert.dom(...).doesNotExist()', () => {
       'Unexpected Parameter: [object HTMLBodyElement]'
     );
 
+    //@ts-ignore -- These assertions are for JavaScript users who don't have type checking
     expect(() => assert.dom(5).doesNotExist()).toThrow('Unexpected Parameter: 5');
+    //@ts-ignore
     expect(() => assert.dom(true).doesNotExist()).toThrow('Unexpected Parameter: true');
     expect(() => assert.dom(undefined).doesNotExist()).toThrow('Unexpected Parameter: undefined');
+    //@ts-ignore
     expect(() => assert.dom({}).doesNotExist()).toThrow('Unexpected Parameter: [object Object]');
+    //@ts-ignore
     expect(() => assert.dom(document).doesNotExist()).toThrow(
       'Unexpected Parameter: [object Document]'
     );
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -23,6 +23,10 @@ export interface ExistsOptions {
   count: number;
 }
 
+export interface Dictionary<T> {
+  [key: string]: T;
+}
+
 export default class DOMAssertions {
   constructor(
     private target: string | Element | null,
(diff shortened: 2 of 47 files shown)


lib/assertions.ts
```diff
@@ -1190,7 +1210,7 @@ export default class DOMAssertions {
   /**
    * @private
    */
-  private get targetDescription() {
+  private get targetDescription(): string {
```
LucasHill: this might fix https://github.com/simplabs/qunit-dom/issues/550 but I'm not 100% sure.

LucasHill:
@Turbo87 is there any appetite for this change?

Turbo87:
generally, yes. but I haven't had time yet to review it properly 😞

LucasHill:
Ok thanks!

lib/assertions.ts
```diff
@@ -575,16 +586,20 @@ export default class DOMAssertions {
    *
    * @see {@link #hasClass}
    */
-  hasPseudoElementStyle(selector: string | null, expected: object, message?: string): void {
+  hasPseudoElementStyle(
+    selector: string | null,
+    expected: { [key: string]: any },
```
scalvert: Can we define a more meaningful type for this? I think this is loosely a dictionary, correct?

```ts
interface Dictionary<T> {
  [key: string]: T;
}
```

scalvert:
@LucasHill thanks for doing this!

lib/assertions.ts
LucasHill: yes that is a good improvment

LucasHill:
@scalvert updated! thanks for the feedback

lib/__tests__/does-not-exist.ts
```diff
@@ -76,10 +76,14 @@ describe('assert.dom(...).doesNotExist()', () => {
       'Unexpected Parameter: [object HTMLBodyElement]'
     );
 
+    //@ts-ignore -- These assertions are for JavaScript users who don't have type checking
```
scalvert: Just a thought about this, but I wonder if it's better to extract these tests out to a js file rather than have to ignore them, when they're clearly aimed at testing the looser allowances of js. 

This would have to be repeated for all below, thought we _could_ put these in a separate subdirectory in tests.

What do you think?

lib/assertions.ts
```diff
@@ -636,16 +655,21 @@ export default class DOMAssertions {
...
 
     let computedStyle = window.getComputedStyle(element, selector);
-    let expectedProperties = Object.keys(expected);
+
+    let expectedProperties = Object.keys(expected) as [keyof CSSStyleDeclaration];
     let result = expectedProperties.some(
       property => computedStyle[property] !== expected[property]
     );
-    let actual = {};
+    let actual: { [key: string]: any } = {};
```
scalvert: This would become `Dictionary<any>`, correct?

lib/assertions.ts
```diff
@@ -636,16 +655,21 @@ export default class DOMAssertions {
    *
    * @see {@link #hasClass}
    */
-  doesNotHavePseudoElementStyle(selector: string | null, expected: object, message: string) {
+  doesNotHavePseudoElementStyle(
+    selector: string | null,
+    expected: { [key: string]: any },
```
scalvert: This would become `Dictionary<any>`, correct?

lib/__tests__/does-not-exist.ts
Turbo87: seems harder to maintain to me, so I'll go ahead an merge as-is :)

PR #591, opened 2020-01-08
Dropping support for node 8

scalvert:
Removes node 8 support, as it was EOL on 12/31.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "typescript": "^3.7.4"
   },
   "engines": {
-    "node": ">= 10.*"
+    "node": "10.* || >= 12.*"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "typescript": "^3.7.4"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": ">= 10.*"
```
Turbo87:
```suggestion
    "node": "10.* || >= 12.*"
```

Node.js 11 is EOL too 😉

scalvert:
It looks like this will fix CI in master.

PR #610, opened 2020-02-02
Add support for assertion chaining

### LINKED ISSUE #177: Make the assertions chainable
What do you think about making it possible to chain assertions? That way you don't have to call `assert.dom()` multiple times (or cache it).

```javascript
assert.dom('.some-class')
  .exists()
  .hasText('Testing')
  .hasAttribute('disabled');
```

Turbo87:
Resolves #177

### DIFF
--- a/lib/__tests__/does-not-exist.ts
+++ b/lib/__tests__/does-not-exist.ts
@@ -84,4 +84,15 @@ describe('assert.dom(...).doesNotExist()', () => {
       'Unexpected Parameter: [object Document]'
     );
   });
+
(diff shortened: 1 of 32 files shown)


Turbo87:
let's do this...! 🤷‍♂ 

and just for the record, I personally would only use this in exceptional cases. it is totally fine not to use this feature. 😉

chriskrycho:
I'm on the same page on all fronts, and appreciate that this *exists* even if I won't usually reach for it! Thanks for adding it!

PR #619, opened 2020-02-10
Add GitHub Actions config

scalvert:
This PR takes a first stab at converting the existing `.travis.yml` to github actions. I ended up splitting the previous configuration into two:

- CI Build: builds, lints and tests
- Publish: builds, tags, and releases

I'm open to changing any of it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  release:
+    name: release
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: yarn install
+      - run: yarn build
+
+      - name: auto-dist-tag
+        run: |
+          yarn global add auto-dist-tag@1
+          auto-dist-tag --write
+
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - 'v*'
+  pull_request: {}
+  schedule:
+    - cron: '0 3 * * *' # daily, at 3am
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        node: ['10', '12']
+
+    name: Tests (Node.js v${{ matrix.node }})
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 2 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: build
+        run: yarn build
+      - name: pre-publish
+        run: |
+          yarn global add auto-dist-tag@1
+          auto-dist-tag --write
+      - name: publish
+        run: yarn publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
scalvert: We'll need to configure this in the GH secrets config.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  release:
+    name: release
+    runs-on: macos-latest
```
Turbo87: why on mac instead of linux?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: install dependencies
+        run: yarn install
+      - name: build
+        run: yarn build
+      - name: pre-publish
+        run: |
+          yarn global add auto-dist-tag@1
+          auto-dist-tag --write
+      - name: publish
+        run: yarn publish
```
Turbo87: I would prefer to use `npm publish` for this. `yarn publish` has never really worked properly for me for some reason 🤷‍♂

.github/workflows/publish.yml
scalvert: Literally no valid reason. I've swapped to `ubuntu-latest`

.github/workflows/publish.yml
scalvert: Ya I've had issues too. Changing it.

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: install dependencies
+        run: yarn install --no-lockfile
+      - name: lint
+        run: yarn lint
+      - name: build
+        run: yarn build
+      - name: test
+        run: |
+          yarn test:coverage --runInBand
+          yarn test:ember
```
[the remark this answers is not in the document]
scalvert: Done. Good suggestion.

.github/workflows/publish.yml
Turbo87: done ✅

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+          node-version: 10
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: yarn install
+      - run: yarn build
+
+      - name: auto-dist-tag
+        run: |
+          yarn global add auto-dist-tag@1
+          auto-dist-tag --write
```
Turbo87: I'm not sure what happened but it looks like this step didn't work: https://github.com/simplabs/qunit-dom/commit/585396cacf0f60a5ae9c9e76d6970b01574056e7/checks?check_suite_id=498999285

PR #634, opened 2020-03-04
Throw an error if an invalid `rootElement` is passed to `assert.dom()`

### LINKED ISSUE #633: `assert.dom('foo', 'bar')` should throw an error
I managed to write an assertion like `assert.dom('foo', 'bar')` today, where I actually meant `assert.dom('foo').hasText('bar')` 😱 

This is clearly a bug, and actually one that we would be able to catch. The `.dom()` function does accept an optional second argument, but if it is not `undefined` then it should be an `Element` (or at least have `querySelector()` and `querySelectorAll()` methods).

I would propose that in the `.dom()` we check that, if a `rootElement` is passed to the function, the `rootElement` has `querySelector()` and `querySelectorAll()` methods and otherwise we throw an error, which would have prevented the issue above.

Turbo87:
Resolves #633

### DIFF
--- a/lib/qunit-dom.ts
+++ b/lib/qunit-dom.ts
@@ -12,6 +12,19 @@ QUnit.assert.dom = function(
   target?: string | Element | null,
   rootElement?: Element
 ): DOMAssertions {
+  if (!isValidRootElement(rootElement)) {
+    throw new Error(`${rootElement} is not a valid root element`);
+  }
+
   rootElement = rootElement || this.dom.rootElement || document;
   return new DOMAssertions(target || rootElement, rootElement, this);
 };
+
+function isValidRootElement(element: any): element is Element {
+  return (
+    !element ||
+    (typeof element === 'object' &&
+      typeof element.querySelector === 'function' &&
+      typeof element.querySelectorAll === 'function')
+  );
+}
--- a/tests/acceptance/qunit-dom-test.js
+++ b/tests/acceptance/qunit-dom-test.js
@@ -39,5 +39,7 @@ module('Acceptance | qunit-dom', function(hooks) {
      * See this: https://github.com/jsdom/jsdom/issues/1928
      */
     assert.dom('#with-pseudo-element').hasPseudoElementStyle(':after', { content: '";"' });
+
+    assert.throws(() => assert.dom('foo', 'bar'), /bar is not a valid root element/);
   });
 });


lib/qunit-dom.ts
```diff
@@ -12,6 +12,19 @@ QUnit.assert.dom = function(
   target?: string | Element | null,
   rootElement?: Element
```
marcoow: I think you'd want to change the type here to something like

```ts
rootElement?: QuerySelectable
…
interface QuerySelectable {
  querySelector(selector: string): Element,
  querySelectorAll(selectors: string): NodeList
}
```

Turbo87: I **do** expect an `Element`, but since the API can be used from JS I can't be sure of it. checking for the two methods seems like a good enough method for us to determine if the passed in thing is an `Element` or not.

Turbo87: the reason why I wouldn't want to touch the function signature is that it will cause changes for users that actually do use this library with TypeScript.

marcoow: An interface like the above would make the signature wider actually so should not result in any problems for TS users. I'd argue that if you want to keep the API as it is you should check for an `Element` to be passed – I can't see a reason for deliberately making it so that the signature and code are out of sync.

Turbo87: > I'd argue that if you want to keep the API as it is you should check for an Element to be passed

how?

marcoow: something like https://lodash.com/docs#isElement ?

marcoow: IMO the other option would be better though

scalvert:
Should we change the label to bug instead, based on the issue identifying this as a bug?

Turbo87:
> based on the issue identifying this as a bug?

I would consider the `dom('foo', 'bar')` usage itself a bug, but I don't think it's necessarily a bug that we don't catch this issue early 😉

lib/qunit-dom.ts
Turbo87: how is https://lodash.com/docs#isElement different/better than the `isValidRootElement()` function in this PR? they're essentially both just guessing, but `isValidRootElement()` at least checks for the things that we actually use.

marcoow: I don't know the details of https://lodash.com/docs#isElement. What I'm basically saying is I think the shape-checking makes sense anyway and I'd reflect that in the TS signature as I cannot see any downsides of doing that and also no upsides of not doing it.

Turbo87: > no upsides of not doing it.

I want people to only pass in `Elements` and nothing else. Also, I think `Element` is common terminology and relatively easy to understand, while `QuerySelectable` is something custom that people would have to look up in our type defs.

also the `DomAssertions` class expects `Element | Document`, so if we go with `QuerySelectable` we would either have to cast it to `Element | Document` again, or we'd have to adjust the class to accept `QuerySelectable` instead too. In that case though we'd run into the problem that the signature of the `querySelect()` functions in `QuerySelectable` is not quite the same as in the `ParentNode` interface, which could then cause issues in our TypeScript code.

scalvert:
Fair. Seems fine.

lib/qunit-dom.ts
marcoow: Too much bikeshedding for me - I’m out

PR #644, opened 2020-03-21
add support for hasCount

### LINKED ISSUE #456: hasCount assertion
I find myself that I am using `exists({ count: 4 })` quite often and I was thinking that a shorthand will be useful - `hasCount`. Some other name might do, but this is more or less in line with `exists`.

```javascript
assert.dom('[data-test-post]').hasCount(3); // I find this easier to write and read
assert.dom('[data-test-post]').exists({ count: 3 });
```

I can create a PR if you think that this is a good idea as well.

mum-never-proud:
Resolves #456

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ High Level DOM Assertions for [QUnit](https://qunitjs.com/)
 assert.dom('h1').exists();
 assert.dom('h1').hasClass('title');
 assert.dom('h1').hasText('Welcome to Ember, John Doe!');
+assert.dom('h1').hasCount(2);
 
 assert.dom('input').isFocused();
 assert.dom('input').hasValue(/.+ Doe/);
--- a/lib/__tests__/has-count.ts
+++ b/lib/__tests__/has-count.ts
@@ -1,0 +1,69 @@
+import TestAssertions from '../helpers/test-assertions';
+
+describe('assert.dom(...).hasCount()', () => {
+  let assert: TestAssertions;
+
+  beforeEach(() => {
+    assert = new TestAssertions();
+  });
+
+  test('succeeds if element exists N times', () => {
+    document.body.innerHTML = '<div></div>'.repeat(3);
+
+    assert.dom('div').hasCount(3);
+
+    expect(assert.results).toEqual([
+      {
+        actual: 'Element div exists 3 times',
+        expected: 'Element div exists 3 times',
+        message: 'Element div exists 3 times',
+        result: true,
+      },
+    ]);
+  });
+
+  test('fails if element exists less than N times', () => {
+    document.body.innerHTML = '<div></div>'.repeat(3);
+
+    assert.dom('div').hasCount(5);
+
+    expect(assert.results).toEqual([
+      {
+        actual: 'Element div exists 3 times',
+        expected: 'Element div exists 5 times',
+        message: 'Element div exists 5 times',
+        result: false,
+      },
+    ]);
+  });
+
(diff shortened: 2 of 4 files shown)


mum-never-proud:
is there are a problem with action flow configuration, looking at the history it's failing since the last build

Turbo87:
see https://github.com/simplabs/qunit-dom/issues/456#issuecomment-524814695

mum-never-proud:
agreed with `exists({ count: 0 })` but anything greater than 0 like  `exists({ count: 5 })`

I find `hasCount(5)` handy it saves me a few strokes and its much cleaner than the former one

@Turbo87

Turbo87:
> and its much cleaner than the former one

I disagree. "[data-test-row] has count 5" sounds weird to me, while "[data-test-row] exists 5 times" sounds much more like readable English. Also, `hasCount()` could just as well mean that the selected element has 5 child elements. I don't think the name is meaningful enough to justify adding the shortcut method. 🤷‍♂

mum-never-proud:
hmm makes sense, but I think there still must be a simplified version of the current implementation 

for starters here are few suggestions

- `existsCount`
- `existsTimes`

or we can have a new assertion HOF `count` that returns a list a functions

- `toEqual`
- `lessThan`
- `greaterThan` etc.


your thoughts @Turbo87

Turbo87:
I don't see how those are simpler... 😅

mum-never-proud:
https://github.com/simplabs/qunit-dom/issues/456#issuecomment-524816171

like this user said it will be visible In autocomplete, we definitely need it, though with a different name or implementation

but I don't think everyone would find `exists({ count: 0 })` convenient

I find it more convenient when autocomplete suggests and supplying a number instead of the current one we can provide as an alias to this

cc: @Turbo87

Turbo87:
> but I don't think everyone would find `exists({ count: 0 })` convenient

that's why `doesNotExist()` can be used too

> I find it more convenient when autocomplete suggests and supplying a number instead of the current one we can provide as an alias to this

autocomplete does suggest `exists()` and if you have a proper autocomplete implementation it should also allow you to autocomplete the `count` option that is specified in the type definition for the `exists()` assertion.

mum-never-proud:
why can't we have a function say

- `existOnce`
- `existTwice`
- `existThrice`
- `existTimes` 

I just wanna understand why you feel these unnecessary? consider this my last question:)

@Turbo87

mupkoo:
I agree that `hasCount` does not sound really nice, but I do think that it will be helpful to have a shorthand for the `exists({ count })`

scalvert:
My take on this is that there are some good assertion primitives here that can be used quite easily and effectively. Adding sugar syntax, while may seem appealing, can sometimes clutter APIs that are otherwise straightforward. 

I can see the appeal, but with correct autocomplete in your editor it's quite easy to type a few extra strokes, but get the clarity from the failed assertion message that you currently do from the use of `exists({ count: n })`.

mum-never-proud:
I agree that with correct autocomplete setup we don't need syntactic sugar using `tabnine` which is powered by AI suggests this but we can't expect everyone to have one.

In my situation, I can have `tabnine` in my PC but in office laptop the antivirus blocks `tabnine` Idk why it's considered as malware but there's were these syntactic sugars come handy in my opinion

I get clarity from `exists({ count: n })`, with a lot of assertions in a file it cumbersome sometimes

so my suggestion is to club and make it as `existsCount(n)`

Issue #736, opened 2020-06-15
Account for null element passed in


patsy-issa:
This is step 1 in an effort to remove potential false positives, as it stands calling `assert.dom().exists()` and `assert.dom(find(nonExistingElement)).exists()` both default to `assert.dom(rootElement)`.

I propose that in the case of a null argument being passed in we can log a warning to the user and recommend the usage of a unique identifier to be used as `assert.dom('unique-identifier')` as the other usage may lead to a false positive.

Turbo87:
I'm not sure about the warning. We should either throw a hard error if `undefined/null` is passed in, or make the function that looks up the element from the passed in selector handles that value correctly (which I think it does already).

patsy-issa:
I m good with an error 👍

PR #746, opened 2020-06-25
Account for null element

patsy-issa:
Adds null `target|element` detection and error. refs #736

### DIFF
--- a/lib/__tests__/does-not-have-tagname.ts
+++ b/lib/__tests__/does-not-have-tagname.ts
@@ -59,14 +59,9 @@ describe('assert.dom(...).doesNotHaveTagName()', () => {
     });
 
     test('fails for missing element', () => {
-      assert.dom(null).doesNotHaveTagName('div');
-
-      expect(assert.results).toEqual([
-        {
-          message: 'Element <unknown> should exist',
-          result: false,
-        },
-      ]);
+      expect(() => assert.dom(null).doesNotHaveTagName('div')).toThrow(
+        'Null target or element was passed'
+      );
     });
   });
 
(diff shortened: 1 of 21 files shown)


lib/__tests__/does-not-have-tagname.ts
```diff
@@ -59,14 +59,9 @@ describe('assert.dom(...).doesNotHaveTagName()', () => {
...
-
-      expect(assert.results).toEqual([
-        {
-          message: 'Element <unknown> should exist',
-          result: false,
-        },
-      ]);
+      expect(() => assert.dom(null).doesNotHaveTagName('div')).toThrow(
+        'Null target or element was passed'
+      );
```
Turbo87, line 64: hmm... why change the existing support for `null` that we already have? 🤔

patsy-issa: `assert.dom(null)` will be throwing the error before `doesNotHaveTagName` gets the chance ^^

lib/__tests__/does-not-have-tagname.ts
```diff
@@ -100,17 +95,6 @@ describe('assert.dom(...).doesNotHaveTagName()', () => {
         },
       ]);
     });
-
-    test('fails for missing element', () => {
-      assert.dom('#missing').doesNotHaveTagName('div');
```
patsy-issa, line 105: Ah I see these should remain, tunnel vision 🤦

lib/__tests__/does-not-have-tagname.ts
patsy-issa: @Turbo87 restored the valid missing element assertions

PR #772, opened 2020-07-30
Fix `assert.dom(null).exists()` case

### LINKED ISSUE #736: Account for null element passed in

Turbo87:
https://github.com/simplabs/qunit-dom/pull/60 unintentionally broke the above case. This commit fixes the implementation by only defaulting to the `rootElement` as `target` when the user hasn't provided any arguments at all.

Resolves #736, Closes #746

/cc @patsy-issa

### DIFF
--- a/lib/qunit-dom.ts
+++ b/lib/qunit-dom.ts
@@ -17,7 +17,12 @@ QUnit.assert.dom = function (
   }
 
   rootElement = rootElement || this.dom.rootElement || document;
-  return new DOMAssertions(target || rootElement, rootElement, this);
+
+  if (arguments.length === 0) {
+    target = rootElement;
+  }
+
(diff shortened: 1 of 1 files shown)


patsy-issa:
https://github.com/simplabs/qunit-dom/pull/746 ;)

Turbo87:
@patsy-issa I'm aware, this PR closes #746, as written in the PR description

PR #773, opened 2020-07-31
Refactor CSS Style Declaration types to support latest TypeScript

BarryThePenguin:
I saw a note in your Chagelog about #757 

This PR is an attempt at solving any issues caused by the latest version of TypeScript 🙈 

Happy to make any changes needed 👋

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -24,9 +24,9 @@ export interface ExistsOptions {
   count: number;
 }
 
-export interface Dictionary<T> {
-  [key: string]: T;
-}
+type CSSStyleDeclarationProperty = keyof CSSStyleDeclaration;
+
+type ActualCSSStyleDeclaration = Partial<Record<CSSStyleDeclarationProperty, unknown>>;
 
 export default class DOMAssertions {
   constructor(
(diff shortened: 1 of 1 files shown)


lib/assertions.ts
```diff
@@ -24,9 +24,9 @@ export interface ExistsOptions {
   count: number;
 }
 
-export interface Dictionary<T> {
-  [key: string]: T;
-}
+type CSSStyleDeclarationProperty = keyof CSSStyleDeclaration;
+
+type ActualCSSStyleDeclaration = Partial<Record<CSSStyleDeclarationProperty, unknown>>;
```
BarryThePenguin, line 29: `CSSStyleDeclaration` requires all properties to be defined

Using `Partial` allows the assertions to define a subset of `CSSStyleDeclaration` properties

Using `unknown` here allows us to get around the issue of `CSSStyleDeclaration` having properties like [`readonly length: number;`](https://github.com/microsoft/TypeScript/blob/d3877d294cb2675381ec00e21251ff533d1c71dc/lib/lib.dom.d.ts#L2929) and [`readonly parentRule: CSSRule | null;`](https://github.com/microsoft/TypeScript/blob/d3877d294cb2675381ec00e21251ff533d1c71dc/lib/lib.dom.d.ts#L2999)

lib/assertions.ts
```diff
@@ -730,14 +730,14 @@ export default class DOMAssertions {
...
     selector: string | null,
-    expected: Dictionary<string>,
+    expected: Record<string, string>,
     message?: string
   ): DOMAssertions {
     let element = this.findTargetElement();
     if (!element) return this;
 
     let computedStyle = window.getComputedStyle(element, selector);
-    let expectedProperties = Object.keys(expected) as [keyof CSSStyleDeclaration];
```
BarryThePenguin, line 740: `[keyof CSSStyleDeclaration]` indicates an array with a single item of `keyof CSSStyleDeclaration`

Is that deliberate?

PR #782, opened 2020-08-09
Fix typing for `hasProperty` assertion

### LINKED ISSUE #781: Incorrect typing for hasProperty assertion
`hasProperty` assertion is defined as `hasProperty(name: string, value: string | RegExp, message?: string)` but properties, unlike attributes, are not limited to just strings.

I’d like to use `hasProperty` to assert for `checked` property on checkbox input in typescript with the following assertion:
```typescript
assert.dom('input').hasProperty('checked', true);
```
but this results in typescript error.

Asserting for stringified value (`"true"`) is incorrect and results in failed test.

wuron:
Fixes #781

### DIFF
--- a/API.md
+++ b/API.md
@@ -475,7 +475,7 @@ expression.
 #### Parameters
 
 -   `name` **[string][121]** 
--   `value` **([string][121] \| [RegExp][132])** 
+-   `value` **([RegExp][132] | any)** 
 -   `message` **[string][121]?** 
 
 #### Examples
--- a/lib/__tests__/has-property.ts
+++ b/lib/__tests__/has-property.ts
@@ -53,6 +53,48 @@ describe('assert.dom(...).hasProperty()', () => {
     });
   });
 
+  describe('boolean expected', () => {
+    test('succeeds for correct name and value', () => {
+      assert.dom('input').hasProperty('disabled', false);
+      assert.dom(document.querySelector('input')).hasProperty('disabled', false);
+
+      expect(assert.results).toEqual([
+        {
+          actual: 'Element input has property "disabled" with value false',
+          expected: 'Element input has property "disabled" with value false',
+          message: 'Element input has property "disabled" with value false',
+          result: true,
+        },
+        {
+          actual: 'Element input[type="password"] has property "disabled" with value false',
+          expected: 'Element input[type="password"] has property "disabled" with value false',
+          message: 'Element input[type="password"] has property "disabled" with value false',
+          result: true,
+        },
+      ]);
+    });
+
+    test('fails for wrong value', () => {
+      assert.dom('input').hasProperty('disabled', true);
+      assert.dom(document.querySelector('input')).hasProperty('disabled', true);
+
+      expect(assert.results).toEqual([
+        {
+          actual: 'Element input has property "disabled" with value false',
+          expected: 'Element input has property "disabled" with value true',
+          message: 'Element input has property "disabled" with value true',
+          result: false,
+        },
+        {
+          actual: 'Element input[type="password"] has property "disabled" with value false',
+          expected: 'Element input[type="password"] has property "disabled" with value true',
+          message: 'Element input[type="password"] has property "disabled" with value true',
+          result: false,
+        },
+      ]);
+    });
+  });
+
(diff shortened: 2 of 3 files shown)


API.md
```diff
@@ -475,7 +475,7 @@ expression.
 #### Parameters
 
 -   `name` **[string][121]** 
--   `value` **([string][121] \| [RegExp][132])** 
+-   `value` **([RegExp][132] \| any)**
```
Turbo87, line 478: this file is automatically generated from the doc comments. make sure to change the the type in the doc comment, and then regenerate it, instead of only changing it here :)

wuron: Oops. :slightly_smiling_face: 
Updated

lib/__tests__/has-property.ts
```diff
@@ -8,45 +8,45 @@ describe('assert.dom(...).hasProperty()', () => {
   beforeEach(() => {
     assert = new TestAssertions();
 
-    document.body.innerHTML = '<input type="password">';
+    document.body.innerHTML = '<input type="checkbox">';
```
Turbo87, line 11: any reason for replacing the existing test instead of adding a new one?

wuron: I thought this way it would be more universal. But I can move it into a separate test if that’s preferred.

wuron:
So what do we need to merge this?

lib/__tests__/has-property.ts
Turbo87: > But I can move it into a separate test if that’s preferred.

yes, I think that would be better to ensure that the existing behavior doesn't break :)

wuron: ok, updated.

lib/assertions.ts
```diff
@@ -520,15 +520,15 @@ export default class DOMAssertions {
...
+   * @param {RegExp|any} value
    * @param {string?} message
    *
    * @example
    * assert.dom('input.password-input').hasProperty('type', 'password');
    *
    * @see {@link #doesNotHaveProperty}
    */
-  hasProperty(name: string, value: string | RegExp, message?: string): DOMAssertions {
+  hasProperty(name: string, value: unknown, message?: string): DOMAssertions {
```
Turbo87, line 531: why are you using `unknown` here instead of `RegExp | any` as specified above?

wuron: `RegExp | any` doesn’t make any sense in typescript, it will still evaluate as `any`. And `unknown` is safer then `any` because it doesn’t imply any knowledge about type.

As for JSDoc annotation there’s no `unknown` type, so `any` is the closest alternative. And it’s still helpful to mention `RegExp` because it has special treatment (match instead of just equality check).

chriskrycho: @wuron is correct: types in TS are *sets*, and since `any` is already a superset of everything else, having `T | any`, for *any* type, is always exactly identical to just having `any`.

chriskrycho: Addendum: you can actually use `unknown` in JSDoc and *most* tools will just show "unknown" there, and anything TS-aware will actually correctly interpret it.

wuron: > you can actually use unknown in JSDoc and most tools will just show "unknown" there, and anything TS-aware will actually correctly interpret it.

If that’s preferred I can set it to `RegExp | unknown` in JSDoc. But does it really change anything? JSDoc definitions are only used as documentation: to generate `API.md` file, or to hint types for non-TS environments. In both cases we want to communicate that this method accepts any value, hence `any` works just fine.

In TS definitions we shouldn’t use `any`, because it’s compatible with all types and allows us to do something like this without any warnings:
```ts
function hasProperty(param: any) {
  param.callRandomMethod();
  const a: boolean = param;
  const b: string = param;
}
```
Parameter with type `unknown` is much safer, and it still allows to pass any value.

chriskrycho:
@Turbo87 would you mind cutting a release which includes this? 🙏 Would love to take advantage of it in a small library I'm open-sourcing!

Turbo87:
v1.5.0 should be out in a couple of minutes once CI is done :)

chriskrycho:
Thank you! Really appreciate it!

PR #816, opened 2020-09-17
Add compatibility with ember-qunit@5

### LINKED ISSUE #812: Compatibility with `ember-qunit@5`
`ember-qunit@5` is (finally 😩) moving along, it brings a few major changes:

* Drop support for older Node versions
* Remove usage of APIs that are not Embroider compatible
  * `{{content-for` usage has been replaced by having the `tests/index.html` directly include our expected DOM content
  * `app.import` has been removed in favor of leveraging imports along with ember-auto-import / Embroider to process them
* Drop direct dependency on `qunit` and `@ember/test-helpers`; moving them to be `peerDependencies` (along with some helpful validations and error messages to help guide folks down the path)

The combination of these things results in a somewhat fundamental problem: tooling can no longer depend on the `QUnit` global to be defined. This is the primary issue that we need to determine how to resolve for `qunit-dom`.

Specifically, at the moment, we directly `app.import` into the `test-support.js` file:

https://github.com/simplabs/qunit-dom/blob/258f8b3a5b0d6f09364d26fec66d454bf3d471e4/index.js#L12-L22

Up until now, this has worked just fine. However, since `QUnit` itself is not directly "splatted" into the `test-support.js` file (it is within a module instead) we are no longer guaranteed that `window.QUnit` will be defined when we call `QUnit.assert.dom = ...` here:

https://github.com/simplabs/qunit-dom/blob/258f8b3a5b0d6f09364d26fec66d454bf3d471e4/lib/qunit-dom.ts#L11

rwjblue:
In order to add support for `ember-qunit@5` the primary Ember project usage pattern changes a bit:

```js
// tests/test-helper.js
import { setup } from 'qunit-dom';
//...snip...
setup(QUnit.assert);
```

No change is required for ember-qunit < 5 (this should not be a breaking change).

Closes #812


scalvert:
Looks good so far. Thanks for implementing explicit calls vs. side-effect imports!

scalvert:
What's the status of this PR, @rwjblue? Can we publish for review, or are you still poking at it?

rwjblue:
> What's the status of this PR, @rwjblue? Can we publish for review, or are you still poking at it?

I still have a few TODO's that need to be completed. 

> Also, publishing ember-qunit, @ember/test-helpers, and others will require a bit of coordination. Do we have a parent issue that determines the correct order of rollout for all these releases, other than Rob's head? :)

Good idea! Created https://github.com/ember-cli/ember-cli/issues/9335 with the TODOs.

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "target": "es5"
   },
   "include": [
-    "lib/**/*"
+    "lib/**/*",
+    "types/**/*"
```
rwjblue, line 13: Can be removed once we land #820

scalvert:
Just did a rebase to fix merge conflicts.

rwjblue:
https://github.com/rwjblue/--ember-qunit-5-sample-app is a demo app that uses the "whole new setup" and is passing.

scalvert:
In terms of the API we expose to users, I'd prefer that we use the pattern that requires the hosting project to pass in `QUnit` to extend.

```js
import { setup } from "qunit-dom"; 

// ...

setup(QUnit);
```

This will not require any special configuration from `qunit-dom` itself, in addition to not making any assumptions about the consumer's configuration of their project. 

Thoughts?

rwjblue:
> In terms of the API we expose to users, I'd prefer that we use the pattern that requires the hosting project to pass in QUnit to extend.

Ya, agreed. That's what the current implementation does (though I need to push a commit to the demo repo above with that change in it).

scalvert:
In the PR description, these two are the same thing, no?
> Document a new API (still somewhat TBD, but probably import "qunit-dom" or import { setup } from "qunit-dom"; setup(QUnit);
> Decide if setup should accept QUnit argument or directly import it?

scalvert:
@rwjblue I updated the description to better reflect the actual final work for this PR.

scalvert:
@Turbo87 I think we've managed to get this into a decent state. The only thing we haven't done here is add automated tests that test the Ember variant with `ember-qunit` > 5, but @rwjblue has a repo that he's been using to manually test. How do we think we should handle this?

.github/workflows/ci.yml
```diff
@@ -29,3 +29,16 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+
+      - run: yarn install --frozen-lockfile
+      - run: yarn test:compatibility
```
Turbo87, line 44: I think I'd prefer to use a `matrix` "strategy" here and have dedicated jobs for each of the ember-try scenarios

index.js
```diff
@@ -7,26 +7,50 @@ const MergeTrees = require('broccoli-merge-trees');
...
+        getDestinationPath() {
+          return 'qunit-dom/index.js';
+        },
+      });
+    } else {
+      let scopedInputTree = new Funnel(`${__dirname}/dist/addon-test-support`, {
+        destDir: 'qunit-dom',
+      });
+
+      return this.preprocessJs(scopedInputTree, '/', this.name, {
```
Turbo87, line 50: what is the purpose of this `this.preprocessJs()` call? might be good to add a code comment :)

ember-cli-build.js
```diff
@@ -8,6 +8,17 @@ module.exports = function (defaults) {
     // Add options here
   });
 
+  if ('@embroider/webpack' in app.dependencies()) {
```
Turbo87, line 11: I'm wondering if it would be better to use a `USE_EMBROIDER` environment variable for this and unconditionally add the dev deps to package.json, instead of using ember-try scenarios. That would allow dependabot to update them, and would not immediately break our CI if embroider releases a version that is no longer compatible.

package.json
```diff
@@ -39,10 +40,14 @@
   },
   "dependencies": {
     "broccoli-funnel": "^3.0.3",
-    "broccoli-merge-trees": "^4.2.0"
+    "broccoli-merge-trees": "^4.2.0",
+    "ember-auto-import": "^1.6.0",
```
Turbo87, line 44: seems a bit unfortunate that any non-ember user will now have to install webpack, but I don't know how to solve this any other way either. having a `ember-qunit-dom` package just to integrate things isn't really great either.

.github/workflows/ci.yml
rwjblue: Sounds good. I was on the fence about this.

package.json
rwjblue: Now that I have the try scenarios setup, I'll try a few other permutations to see if we can avoid it.

rwjblue: just confirmed, and removed

ember-cli-build.js
rwjblue: Done!

index.js
rwjblue: Updated!

.github/workflows/ci.yml
rwjblue: Done!

.github/workflows/ci.yml
```diff
@@ -29,3 +29,23 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+
+      - run: yarn install --frozen-lockfile
+      - run: node_modules/.bin/ember try:one --config-path tests/dummy/config/ember-try.js ember-qunit-${{ matrix.ember-qunit-version}}
+        env:
+          USE_EMBROIDER: "${{ matrix.ember-qunit-version }}"
```
Turbo87, line 51: I assume this is a typo? 😅

package.json
```diff
@@ -28,6 +28,7 @@
     "test": "jest",
     "test:coverage": "jest --coverage",
     "test:ember": "ember test",
+    "test:compatibility": "ember try:each --config-path tests/dummy/config/ember-try.js",
```
Turbo87, line 31:
```suggestion
```

I guess we don't need this anymore :)

tests/dummy/config/ember-try.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+/* eslint-env node */
+
+module.exports = async function () {
+  return {
+    useYarn: true,
+    scenarios: [
+      {
+        // this is currently the version in package.json, so no dependnecy changes are needed
```
Turbo87, line 9:
```suggestion
        // this is currently the version in package.json, so no dependency changes are needed
```

Turbo87 (approved):
LGTM after these minor comments are resolved. nice work! 👍

PR #820, opened 2020-09-23
Update @types/qunit to latest.

rwjblue:
The latest version includes the types for the `qunit` module. Prior versions relied on `@types/ember-qunit` to add the `qunit` module, but since this package doesn't use `ember-qunit`'s types we could not import from `qunit` (without adding custom types).


Turbo87 (approved):
thanks 👍

PR #823, opened 2020-09-24
Do not default `target` to `document`.

rwjblue:
The prior behavior was pretty broken, in the case where `assert.dom()` is called (no arguments) `rootElement` will be `document` for non-Ember users (because `rootElement` falls back to `document`), but our types do not allow `document` for the `target` property of `DOMAssertions`.

It isn't _just_ a types issue though, things like `findElement`/`findElements` actually error if `this.target` is not `null`, a string, or `instanceof Element` (which `document` is not)!

This changes the defaulting to set `target` to the `rootElement` only when it is valid (when its an `Element` instance).

### DIFF
--- a/lib/qunit-dom.ts
+++ b/lib/qunit-dom.ts
@@ -19,7 +19,7 @@ QUnit.assert.dom = function (
   rootElement = rootElement || this.dom.rootElement || document;
 
   if (arguments.length === 0) {
-    target = rootElement;
+    target = rootElement instanceof Element ? rootElement : null;
   }
 
   return new DOMAssertions(target, rootElement, this);


lib/qunit-dom.ts
```diff
@@ -19,7 +19,7 @@ QUnit.assert.dom = function (
   rootElement = rootElement || this.dom.rootElement || document;
 
   if (arguments.length === 0) {
-    target = rootElement;
+    target = rootElement instanceof Element ? rootElement : null;
```
rwjblue, line 22: This is basically the same check that `findElement`/`findElements` does internally, and will ensure that `ember-qunit` users are unaffected by these changes (since Ember users have a customized `QUnit.assert.dom.rootElement` that is the `#ember-testing` div).

scalvert:
Thanks for finding this, @rwjblue. There's been a few jagged edges WRT `target` before.

This is separate from the stuff we were looking at this morning, correct? Specifically, the setting of root element via other means than a getter on `assert.dom`.

rwjblue:
> This is separate from the stuff we were looking at this morning, correct? Specifically, the setting of root element via other means than a getter on assert.dom.

Yep, I was working on that and getting type failures (once I actually told the `assert.dom` function what its `this` was, it was like "hey, can't do that")...

scalvert:
Got it. I figured this shook out from that work.

scalvert (approved):

Turbo87:
should this be considered a breaking change?

rwjblue:
> should this be considered a breaking change?

@Turbo87 - IMHO, no. As I mentioned above, if you actually did end up falling back to `document` before it would always error. I think the only reason this wasn't caught was that for the most part, folks in the Ember community wouldn't hit this because of the `rootElement` override there.

Turbo87:
hmm, I'm still somewhat surprised by this because the Jest-based tests are using the regular document 🤔

rwjblue:
> hmm, I'm still somewhat surprised by this because the Jest-based tests are using the regular document 🤔

Jest doesn't use real DOM AFAICT, maybe an issue there? Or perhaps you never actually hit the fallback (e.g. not calling `assert.dom()` without any args?)?

rwjblue:
When I search the repo for `dom()` the only matches I find are in `tests/acceptance/qunit-dom-test.js`.

Turbo87:
hmm, I see. though I'm still not sure how this fixes the problem. if we pass `null` instead, wouldn't that also cause issues?

scalvert:
I just saw this comment now, @Turbo87. Sorry to not respond sooner.

I think this fix works because passing no args [defaults the target](https://github.com/simplabs/qunit-dom/blob/master/lib/qunit-dom.ts#L21), as you know, and passing null works because subsequent [`findElement`](https://github.com/simplabs/qunit-dom/blob/83c9cfd8811bb13bc783d7fecd312c72dba1d7a9/lib/assertions.ts#L1370) and [`findElements`](https://github.com/simplabs/qunit-dom/blob/83c9cfd8811bb13bc783d7fecd312c72dba1d7a9/lib/assertions.ts#L1388) calls both handle the case when we pass null.

Let me know if this answers the question.

PR #847, opened 2020-10-19
Sort `package.json` fields

rwjblue:
The result of running `npx sort-package-json` (no modifications otherwise).


scalvert (approved):
Context on this is we want to get a stable package.json sort order for subsequent changes.

Turbo87 (approved):
thanks :)

PR #894, opened 2020-11-20
Move API.md link underneath a Documentation header

locks:


Turbo87:
I don't see the point of this. the whole README file is "Documentation" and IMHO the API of the library is how you use the library, so having it under "Usage" seems fitting 🤔

locks:
It makes it easier for me to scan for the headers to find the link to the documentation than to read the content of the Usage header.
I think my viewpoint makes more sense if I tell you I would consider the "Ember projects using ember-qunit v5.x and above" as part of Usage and not Install. Running the yarn/npm command install the dependency, but to actually use it, and specifically in an `ember-qunit` project, you need this extra instruction, just like we have the extra instruction for using with TypeScript.

scalvert:
Though it is a bit easier to find, the README is very small and isn't too tricky to scan. That said, if it helps _anyone_ find the docs better, then maybe it's a reasonable change?

locks:
Let's leave it as is until there's something more significant to add.

PR #983, opened 2021-01-22
doc(release): remove extra 'to'

mrloop:


Turbo87 (approved):
whoops, thanks :)

rwjblue (commented):
Thanks for cc'ing @Turbo87, I'll double check the template to make sure I don't have this typo there too.

PR #1139, opened 2021-05-31
Dependencies cleanup

izelnakri:
- Removed all redundant ember-cli boilerplate code.
- Removed github actions ember compatibility CI that depends on ember-cli. This project source code should target browser, not ember-cli. Ember projects can simply use ember-auto-import.
- Replace ember acceptance with qunitx/browser compliant code.
- Removed redundant dev dependencies(babel, rollup)
- Moved dev dependency config files to package.json

I didn't touch most of the test suite which is based on jest. As it seems jest runs the code in jsdom environment, which is actually the most unrealistic environment to run the code, thats why there were previously existing runtime errors. Perhaps there is a way to run the tests on puppeteer. You can also ditch jest completely and run everything easily on qunitx now.

I think advantages for not having install dependencies and more realistic test runtimes were self-evident, however I do see some push back here: https://github.com/simplabs/qunit-dom/issues/1135 Thus, I'll continue maintaining a fork and develop this further in different branches in case if this PR doesn't get merged due to some disagreement.


Turbo87:
I would recommend reading through https://simplabs.com/blog/2021/05/26/keeping-a-clean-git-history/ specifically this part:

> Open source projects tend to receive a large number of PRs out of the blue, some of them are great but sometimes you get the dreaded "rewrite all your code" PR. These are often very hard to ever get merged (if you would even want to merge them) because they end up touching too much stuff in the same PR and can make it next to impossible to review properly.

and if you need more information on this topic: 

- https://mtlynch.io/code-review-love/
- https://google.github.io/eng-practices/review/developer/

This PR changes roughly 200 different things, which basically makes it impossible to review and merge. While some of the changes might be useful on their own, I can't simply cherry pick them because they are not done in dedicated commits. Unfortunately that leaves me no other choice but to reject this PR completely, sorry.

izelnakri:
You could squash all commits and git apply and make the commits you like, I cant know what commits you deem ideal and I do find the current commits easy to follow actually.

You reacted as I expected, Im not going to create another PR and didnt need that git lecture thanks.

PR #1194, opened 2021-07-29
feat: add `withAriaSupport` option

buschtoens:
As suggested in https://github.com/simplabs/qunit-dom/issues/915#issuecomment-868330633, this PR adds a `withAriaSupport` option, so that [`aria-*`](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) options are considered.

If not explicitly set, it defaults to `false` for all assertions, except for `isChecked` / `isNotChecked`, where it defaults to `true`. This way this is a non-breaking change.

- [ ] Implementation
  - [x] `isChecked` / `isNotChecked`: [`aria-checked`](https://www.w3.org/TR/wai-aria-1.1/#aria-checked)
  - [x] `isRequired` / `isNotRequired`: [`aria-required`](https://www.w3.org/TR/wai-aria-1.1/#aria-required)
  - [x] `isDisabled` / `isNotDisabled` / `isEnabled`: [`aria-disabled`](https://www.w3.org/TR/wai-aria-1.1/#aria-disabled)
  - [x] `isValid` / `isNotValid`: [`aria-invalid`](https://www.w3.org/TR/wai-aria-1.1/#aria-invalid)
- [ ] Unresovled Questions:
  - [ ] Should `isDisabled` / `isNotDisabled` / `isEnabled` also (optionally?) consider:
    - [`aria-readonly`](https://www.w3.org/TR/wai-aria-1.1/#aria-readonly)
    - [`aria-busy`](https://www.w3.org/TR/wai-aria-1.1/#aria-busy)
    - Should these (only?) be added as extra assertions, like: `isReadonly`, `isBusy`?
  - [ ] Should `isVisible` consider [`aria-hidden`](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden)?
  - [ ] What about these attributes?
    - [`aria-expanded`](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded)
    - [`aria-pressed`](https://www.w3.org/TR/wai-aria-1.1/#aria-pressed)
    - [`aria-selected`](https://www.w3.org/TR/wai-aria-1.1/#aria-selected)
    - [`aria-valuenow`](https://www.w3.org/TR/wai-aria-1.1/#aria-valuenow)
    - [`aria-valuetext`](https://www.w3.org/TR/wai-aria-1.1/#aria-valuetext)
- [ ] Tests
- [ ] Docs

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -20,6 +20,25 @@ export interface AssertionResult {
   message: string;
 }
 
+export interface DOMAssertionOptions {
+  /**
+   * Whether to also consider `aria-*` attributes in semantic assertion helpers.
+   *
+   * {@link https://www.w3.org/TR/wai-aria-1.1/#state_prop_def}
+   *
+   * @see DOMAssertions#isChecked
+   * @see DOMAssertions#isNotChecked
+   * @see DOMAssertions#isRequired
+   * @see DOMAssertions#isNotRequired
+   * @see DOMAssertions#isDisabled
+   * @see DOMAssertions#isNotDisabled
+   * @see DOMAssertions#isEnabled
+   * @see DOMAssertions#isValid
+   * @see DOMAssertions#isNotValid
+   */
+  withAriaSupport?: boolean;
+}
+
 export interface ExistsOptions {
   count: number;
 }
@@ -32,7 +51,8 @@ export default class DOMAssertions {
   constructor(
     private target: string | Element | null,
     private rootElement: Element | Document,
-    private testContext: Assert
+    private testContext: Assert,
+    private options: DOMAssertionOptions = {}
   ) {}
 
   /**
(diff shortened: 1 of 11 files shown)


lib/assertions.ts
```diff
@@ -172,9 +194,10 @@ export default class DOMAssertions {
    * assert.dom('input[type="text"]').isRequired();
    *
    * @see {@link #isNotRequired}
+   * {@link https://www.w3.org/TR/wai-aria-1.1/#aria-required}
    */
-  isRequired(message?: string): DOMAssertions {
-    isRequired.call(this, message);
+  isRequired(message?: string, options?: { withAriaSupport?: boolean }): DOMAssertions {
+    isRequired.call(this, message, options);
```
Turbo87, line 200: I think this is missing the fallback to `this.options.withAriaSupport` if the option is not explicitly specified in the passed in `options` argument

Turbo87: hmm, looks like you've moved that fallback into the functions. I think I'd prefer it if the functions would always get an `options` object with a `withAriaSupport` value and don't have to depend on `this` except for `this.pushResult()`. 🤔

PR #1241, opened 2021-09-17
Fix documentation for pseudo element style

mfeckie:
When using `getComputedStyle` according to the MDN page https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle, we shouldn't pass the `:`.

Have tested this on our local setup and tests pass *without* the ':', but fail with it.

### DIFF
--- a/API.md
+++ b/API.md
@@ -615,7 +615,7 @@ Assert that the pseudo element for `selector` of the [HTMLElement][] has the `ex
 #### Examples
 
 ```javascript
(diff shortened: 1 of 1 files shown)


Turbo87:
this looks like a pretty big diff 😅 

also, the `API.md` file is auto-generated. unless you change it in the corresponding code comment the change will be overridden again soon.

mfeckie:
Huh! Whoops. I only changed 2 characters direct in the GitHub editor. 

Will follow up with something something updating from where it's generated

PR #1295, opened 2021-10-20
Contains classes

NullVoxPopuli:
I found a need for this, as the existing apis felt cumbersome for testing multiple classes.

Thoughts?


NullVoxPopuli:
this is possible now with `hasClass`

https://github.com/mainmatter/qunit-dom/blob/6cb21ddbb143c659e716d0b3024de5d0fd4fd127/packages/qunit-dom/lib/assertions.ts#L722

PR #1379, opened 2021-12-17
Add assertion is-not-overflowing

tylerbecks:
I verified this works by testing against a local codebase with failing and passing test cases.


scalvert:
@tylerbecks thanks for the addition. I think the names don't match what your PR title says. The assertions appear to be `isOverflown` and `isNotOverflown`. These names don't represent correct english, and also are are past tense. It may be better to name them `isOverflowing` and `isNotOverflowing`, both present tense. What do you think?

tylerbecks:
> 

I originally had those names but didn't see any other assertions with the present participle.  But agreed, real engilsh would be best.  Updating!

tylerbecks:
I had to run `yarn upgrade` because `ember test` was failing with this error:

```
yarn run v1.22.17
$ ember test
Environment: test
cleaning up...
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-source]) in @ember/-internals/overrides/index.js

/Users/tbecks/Code/LinkedIn/qunit-dom/@ember/-internals/overrides/index.js: 'loose' mode configuration must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled).


Stack Trace and Error Report: /var/folders/t9/s271r3dn0vq71w83h31q9m680010cz/T/error.dump.db8174e787d8c867f07ef0ae3f0f4dd1.log
error Command failed with exit code 1.
```

scalvert:
Does that mean master isn't passing? 

I'd probably avoid running `yarn upgrade`, as it just does a whole hog upgrade of all your packages at once. At a minimum, I'd do that in a separate PR than this, in order to keep this PR focused. Can we do that instead, separately upgrade the necessary packages?

README.md
```diff
@@ -89,15 +87,14 @@ test('the title is welcoming', function(assert) {
...
-
 ### TypeScript
 
 `qunit-dom` includes type definition files, but the way it extends QUnit means
 that you need import it somewhere so that TS and your editor can pick up the
 types. It is recommended to add the following line to your
 `tests/test-helper.ts` file:
 
-```
+```bash
```
Turbo87, line 97: these changes appear to be unrelated to the proposed feature, no?

lib/assertions.ts
```diff
@@ -1,14 +1,16 @@
...
 import focused from './assertions/focused';
-import notFocused from './assertions/not-focused';
 import isChecked from './assertions/is-checked';
+import isDisabled from './assertions/is-disabled';
 import isNotChecked from './assertions/is-not-checked';
-import isRequired from './assertions/is-required';
+import isNotOverflowing from './assertions/is-not-overflowing';
 import isNotRequired from './assertions/is-not-required';
+import isOverflowing from './assertions/is-overflowing';
+import isRequired from './assertions/is-required';
```
Turbo87, line 9: I agree that ordering these alphabetically probably makes sense, but that is also unrelated to adding the feature here. if you want to refactor things that are unrelated to the feature then I would recommend doing so in dedicated PRs (or at least commits), so that these changes are easier to review.

for more info on these topics have a look at https://mtlynch.io/code-review-love/ :)

yarn.lock
```diff
@@ -0,0 +1,15453 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Turbo87: we've migrated this project over to https://pnpm.io instead of yarn. the yarn lockfile should not be added to the repository anymore to avoid confusion.

tylerbecks: Got it, reverted.

tylerbecks: I ran the upgrade in a seperate PR: https://github.com/simplabs/qunit-dom/pull/1384

lib/assertions.ts
tylerbecks: Thanks for sending this over! I'll undo the sorting here.

README.md
scalvert: +1, I'd remove these.

tylerbecks: Removed

tylerbecks:
@Turbo87 thanks for your review, just updated based on all your comments.

tylerbecks:
Hey everyone 👋 just coming back from vacation and looping back to this.  

I think my previous commit was a no-op and I provided a bit of context in a comment here: https://github.com/simplabs/qunit-dom/pull/1384

@scalvert @Turbo87 I'd love to circle back and get this merged if everything looks good.  

@Turbo87 We had a rocky start but hopefully I still have a chance to make you fall in love with me.

scalvert (approved):

scalvert:
LGTM. Interested in @Turbo87's review.

Turbo87:
looks generally alright, although I'm uncertain how useful this assertion will ultimately be. since it potentially depends on the browser window size and other dynamic factors it seems a bit brittle to me on first glance.

PR #1384, opened 2021-12-20
Run pnpm upgrade

tylerbecks:


scalvert (approved):

Turbo87:
@tylerbecks why did you update these values? this PR does not appear to have any description at all 🤔

tylerbecks:
@Turbo87 I updated the description, this PR is to unblock this one https://github.com/simplabs/qunit-dom/pull/1379

Turbo87:
> I had to run yarn upgrade

but this PR changes exactly nothing about the installed packages... it only updates the dependency requirements, but those were already met by the existing lockfile, which essentially wasn't changed by this PR.

tylerbecks:
@Turbo87 I was experiencing issues running `ember test` and it seemed like it came from some issue with broccoli and maybe babel.  I had a hunch that running `upgrade` would fix it, and it did.  I see your point with `broccoli-funnel` because it only changed a patch version in the lock file. I don't have a good explanation for why this fixed it, but it seemed to fix it.

Turbo87:
> I see your point with `broccoli-funnel` because it only changed a patch version in the lock file.

my point is that this PR did change absolutely nothing in the lockfile, except for the dependency requirements, but not the actual versions that are being installed.

scalvert:
Good call @Turbo87, and my bad for not noticing this (I blanked out on the pnpm-lock).

@tylerbecks - what was the specific issue you were seeing originally?

tylerbecks:
> 

I can't seem to reproduce the issue now traveling back to the commit before this one, but my hypothesis is that this issue was caused by me using `yarn` to install the dependencies

PR #1481, opened 2022-03-02
Fix `hasAria()` docs and add `hasNoAria()` and `lacksAria()` aliases

jamescdavis:
This:
- updates the link in `hasAria()`'s doc block from `hasNoAria()` to `doesNotHaveAria()`
- adds `hasNoAria()` and `lacksAria()` as aliases for `doesNotHaveAria()` to be consistent with `hasNoAttribute()` and `lacksAttribute()` aliases for `doesNotHaveAttribute()` (and other doesNotHave assertions)
- updates the example in `doesNotHaveAttribute()` to use `doesNotHaveAttribute()` instead of its alias `hasNoAttribute()`

Note: I can't get `pnpm run docs` to succeed so this PR doesn't update `API.md`

### DIFF
--- a/lib/__tests__/has-aria.ts
+++ b/lib/__tests__/has-aria.ts
@@ -15,12 +15,26 @@ describe('assert.dom(...).hasAria()', () => {
   test('smoke test', () => {
     assert.dom('button').hasAria('pressed', 'true');
     assert.dom('button').doesNotHaveAria('hidden');
+    assert.dom('button').hasNoAria('hidden');
+    assert.dom('button').lacksAria('hidden');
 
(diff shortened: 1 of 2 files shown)


jamescdavis:
I think it makes sense to add `hasNoAria()` and `lacksAria()`  as aliases to be consistent with the `hasNoAttribute()` and `lacksAttribute()` aliases for `doesNotHaveAttribute()`.

jamescdavis:
@Turbo87 I've expanded this PR to add those aliases.

Turbo87 (approved):
thanks :)

PR #1674, opened 2022-08-12
Support ESM

NullVoxPopuli:
today, qunit-dom's default path points to the wrong file.

in a real ESM environment, `import * as anything from 'qunit-dom'` points at the ember addon index.js instead of anything useful.

I've been working towards getting v2 addon testing to be much less a pain in the butt, so I'm going full ESM with Vite (which I know will require a bunch of work when I get to dealing with the emberisms), but qnuit-dom should probably be compatible with ESM environments anyway. Here is the patch I made here: https://github.com/NullVoxPopuli/ember-resources/pull/585/files#diff-654ac9908e210063f32fb8d67f4742e747b8b19fd85f087b265fe81147eda554
Some proof: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,11 @@
   "repository": "https://github.com/simplabs/qunit-dom",
   "license": "MIT",
   "author": "simplabs GmbH",
+  "exports": {
+    ".": {
+      "import": "./dist/addon-test-support/index.js"
+    }
+  },
   "types": "dist/qunit-dom.d.ts",
   "directories": {
     "doc": "doc",


Turbo87:
given https://github.com/emberjs/data/issues/8296#issuecomment-1325548485, I'm not sure about merging this 🤔

SergeAstapov:
@Turbo87 do you think it worth converting this to v2 addon to achieve wheat is being asked here?

Turbo87:
depends on what that means 😅

NullVoxPopuli:
@Turbo87 a v1 addon with auto import can support exports.

It's the V1 addon itself that can't have exports. Everything else is 'just handled' (assuming auto import @2 is present)

Converting to a v2 addon is mostly mechanical file moves
- after the tests are split out
- npm packages has compiled assets only
 Stuff like that.

It's an ecosystem requirement to convert all addons to the v2 requirement so that we can one day use modern tooling

NullVoxPopuli:
Gonna close this in favor of converting the library to a v2 addon.

Given how broad the support is for qunit-dom today, some breaking changes will be needed.

PR #1695, opened 2022-09-19
Rebrand simplabs to Mainmatter

BobrImperator:
#1696 should fix the pipeline

PR #1696, opened 2022-09-19
chore: downgrade documentation@13.2.5 to 13.2.4

BobrImperator:
unblocks #1695 
Seems like the `13.2.5` and newer are breaking the docs pipeline.
#1681 also gives different output.


Turbo87:
hmm... weird that https://github.com/simplabs/qunit-dom/pull/1539 was automerged 🤔 

might be worth disabling automerge for the `documentation` package

PR #1791, opened 2022-12-20
Add `HTMLLinkElement` as eligible `.disabled` element

gossi:
When trying to write tests for a `HTMLLinkElement` and check for `.disabled` state, instead of asserting, `qunit-dom` throw an error.

There is a little gotcha for `disabled` on `<link>`.

- As per declarative API ([`<link disabled>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-disabled)), this is experimental, deprecated and advised against its usage (and also does not work in browsers)
- As per imperative API ([`HTMLLinkElement.disabled`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLinkElement#htmllinkelement.disabled)), this is according to MDN allow _and works in browsers_

So, this PR adds `HTMLLinkElement` as eligible element to check this for 😃

### DIFF
--- a/lib/assertions/is-disabled.ts
+++ b/lib/assertions/is-disabled.ts
@@ -12,7 +12,8 @@ export default function isDisabled(message?: string, options: { inverted?: boole
       element instanceof HTMLButtonElement ||
       element instanceof HTMLOptGroupElement ||
       element instanceof HTMLOptionElement ||
-      element instanceof HTMLFieldSetElement
+      element instanceof HTMLFieldSetElement ||
+      element instanceof HTMLLinkElement
     )
   ) {
     throw new TypeError(`Unexpected Element Type: ${element.toString()}`);


Turbo87:
but... why? 😂 

is there a legitimate use case for something like this? 😅

gossi:
In fact yes, there is 😅

When each `<link>` represents a theme and only one can be active, then make the others `.disabled` in which case the browser don't use that CSS. That's the state when CSS is loaded but not in use.

Turbo87:
and why not remove the `link` elements instead? 🤔

gossi:
that's implementation details of my particular library.

The point is, the provided API (`isDisabled()` and `isNotDisabled()`) is here to check for a `disabled` property on any `* extends HTMLElement` that supports it. However, this is subjectively constrained by `qunit-dom` and doesn't match reality. I figured `HTMLLinkElement` is missing from that list and I wouldn't be surprised if there are more.

I see this more of a bugfix, so the API keeps holding up to its promise :)

PR #1801, opened 2022-12-31
Add eslint-plugin-jest

bmish:
Add linter for tests: https://github.com/jest-community/eslint-plugin-jest

### DIFF
--- a/lib/__tests__/does-not-have-attribute.ts
+++ b/lib/__tests__/does-not-have-attribute.ts
@@ -52,7 +52,7 @@ describe('assert.dom(...).doesNotHaveAttribute()', () => {
     ]);
   });
 
-  test('fails for wrong content (string value)', () => {
+  test('fails for wrong content (boolean value)', () => {
     assert.dom('input').doesNotHaveAttribute('required');
     assert.dom(document.querySelector('input')).doesNotHaveAttribute('required');
 
--- a/lib/__tests__/is-visible.ts
+++ b/lib/__tests__/is-visible.ts
@@ -32,21 +32,6 @@ describe('assert.dom(...).isVisible()', () => {
         },
       ]);
     });
-
-    test('fails if element is missing', () => {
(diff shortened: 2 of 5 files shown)


lib/__tests__/does-not-have-attribute.ts
```diff
@@ -52,7 +52,7 @@ describe('assert.dom(...).doesNotHaveAttribute()', () => {
     ]);
   });
 
-  test('fails for wrong content (string value)', () => {
+  test('fails for wrong content (boolean value)', () => {
```
bmish, line 55: Detected by [no-identical-title](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-identical-title.md).

lib/__tests__/is-visible.ts
```diff
@@ -32,21 +32,6 @@ describe('assert.dom(...).isVisible()', () => {
         },
       ]);
     });
-
-    test('fails if element is missing', () => {
```
bmish, line 36: [no-identical-title](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-identical-title.md), was a duplicate test case.

Turbo87: hmm, I think the original intention here was that the first test was the happy path with `assert.dom('h1').isVisible();` 😅

bmish: I see. I just checked and it looks like we can't test the happy path here due to:

```
 * JSDom based tests aren't able to discern visibility as we define it. Specifically,
 * the JSDom tests don't do layouting, therefore calculating `offsetWdith` or `offsetHeight`
 * won't work. As a result, we need to use Ember's test infrastructure to correctly assess
 * visibility, as those tests run in a browser environment.
 *
 * Additional tests cases of isVisible can be found in tests/acceptance/qunit-dom-test.js
```

So removing the test case should be fine.

Turbo87: makes sense, yeah

Issue #1972, opened 2023-06-30
Assert css variable value?


jkeen:
It doesn't seem like qunit-dom has the ability to check the value of a css custom property/variable, like `--color-text`, or whatever. Am I right about that?

I tried using `hasStyle`, `hasAttribute`, and `hasProperty` with no avail. In trying to make some changes for a PR, but it seems ilke the issue might be a jest / node thing?

BobrImperator:
With #1976 merged it might've been fixed, at least for the `hasStyle` test.

Feel free to reopen if it's not fixed.

PR #1976, opened 2023-07-01
Use getProperty for hasStyle assertions so testing can be done against css variables

jkeen:
As described in #1972, qunit-dom doesn't currently have the ability to check css variables. This PR adds that ability.

### DIFF
--- a/lib/assertions.ts
+++ b/lib/assertions.ts
@@ -774,11 +774,13 @@ export default class DOMAssertions {
     }
 
     let result = expectedProperties.every(
-      property => computedStyle[property] === expected[property]
+      property => computedStyle.getPropertyValue(property.toString()) === expected[property]
     );
     let actual: ActualCSSStyleDeclaration = {};
 
-    expectedProperties.forEach(property => (actual[property] = computedStyle[property]));
+    expectedProperties.forEach(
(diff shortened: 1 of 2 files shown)


lib/assertions.ts
```diff
@@ -774,11 +774,13 @@ export default class DOMAssertions {
     }
 
     let result = expectedProperties.every(
-      property => computedStyle[property] === expected[property]
+      property => computedStyle.getPropertyValue(property.toString()) === expected[property]
```
BobrImperator, line 777: A `property` seems to be a string already, is the `toString` method needed?

jkeen: I agree that it's a little silly, but it's this sort of nonsense that typescript is very serious about.

BobrImperator: It seems like this bit is faulty in that case
`let expectedProperties = Object.keys(expected) as CSSStyleDeclarationProperty[];`

Since the `getPropertyValue` method expects a string and not a `CSSStyleDeclarationProperty` I assume.
It's possible that we could get rid of that cast.

Feel free to leave it :+1:

BobrImperator:
@jkeen https://www.npmjs.com/package/qunit-dom/v/3.1.0 released :rocket:

Issue #2022, opened 2023-09-16
Unable to use with Vite in non-ember projects


sukima:
I discovered a bug when following the instructions to install qunit-dom. I seems to expect Vite to include Broccollii with the bundled for browser output. I think this might have to do with the `package.json` not providing a `main` and the default `index.js` is an ember-cli addon module.

The basic premise was I had qunit working and I added qunit-dom like this:

1. Add qunit-dom to `package.json` devDependencies
2. Add `import ‘qunit-dom’;` to my test start up module (after first import of qunit)
3. Run vite (dev server)
4. Load page in browser

The console produces the following error:

```
Uncaught ReferenceError: process is not defined
    at node_modules/quick-temp/node_modules/rimraf/rimraf.js (rimraf.js:23)
    at __require2 (chunk-OROXOI2D.js?v=1ddafd9e:16)
    at node_modules/quick-temp/index.js (index.js:4)
    at __require2 (chunk-OROXOI2D.js?v=1ddafd9e:16)
    at node_modules/broccoli-plugin/dist/read_compat.js (read_compat.ts:4)
    at __require2 (chunk-OROXOI2D.js?v=1ddafd9e:16)
    at node_modules/broccoli-plugin/dist/index.js (index.ts:11)
    at __require2 (chunk-OROXOI2D.js?v=1ddafd9e:16)
    at node_modules/broccoli-funnel/index.js (index.js:6)
    at __require2 (chunk-OROXOI2D.js?v=1ddafd9e:16)
```

## Steps to reproduce

1. Clone https://github.com/sukima/lit-app-starter-kit
2. Install packages `bun i`, `npm i`, `pnpm i` your preference I chose Bun
3. run the `dev` script `bun run dev`
4. Open in browser `http://localhost:3000/tests/`

This is the problematic commit: https://github.com/sukima/lit-app-starter-kit/commit/42220f7d7ab878a73a5dd4c9d3251262f529aae9

You can verify it is qunit-dom by `git checkout HEAD^` and running the steps above.

NullVoxPopuli:
I think defining an exports key should also help resolve, which we'd probably want to align with modern package.jsons (provided it includes default and types)

Additionally, we can specify for v1 addons to not use index.js via specifying ember-addon-main in ember-addon

NullVoxPopuli:
[...]

PR #2027, opened 2023-09-23
Prepare for V2 Addon conversion.

NullVoxPopuli:
Note the diff stats are way more intimudating than this PR actually is.
It's 90% "moves" + some boilerplate.
Very little has actually changed.

----

_The primary purpose of V2 Addon conversions is to get ember apps installing deps, booting and building faster_
(for me, anyway)

----

qunit-dom has never been compatible with modern packagers -- this is due to the issue reported in #2022 partially due to incorrect package.json config.

Since we want to also be good stewards of the ember ecosystem, we don't want to _just_ fix the issue for non-ember users. We want to make sure that ember users don't pay the v1 addon / broccoli infrastructure cost when consuming this library.

Step one: convert to monorepo
 - extracting the tests to their own project
 - ensuring CI is up to date to be ok with the new structure
Step two: convert the addon to v2
  - thankfully the library is already most of the way there -- this step will mostly be package.json tweaks
    - the package cross-compiles to IIFE (I presume for non-ESM CDN usage?) and ESM already.
  - this step will mostly be getting rid of the index.js, and other "emberisms"

 

For additional context on this process, I've written up this guide: https://gist.github.com/NullVoxPopuli/eafc7dad6547de5e730098498b829e1f
- I've skipped the `1. Convert solo-repo to single-package monorepo` step because I know that the MainMatter folks are more up to date with what's going on in the ember ecosystem than most, and the step is very small anyway.

Thankfully, qunit-dom is already using pnpm :tada:


.github/workflows/ci.yml
```diff
@@ -84,6 +84,7 @@ jobs:
 
       - run: pnpm install
       - run: pnpm build
-      - run: pnpm ember try:one ${{ matrix.ember-try-scenario}} -- --config-path tests/dummy/config/ember-try.js
+      - run: pnpm ember try:one ${{ matrix.ember-try-scenario}} --skip-cleanup 
+        working-directory: test-app
```
BobrImperator, line 88: I think we should do `pnpm --filter test-app try:one ${{ matrix.ember-try-scenario}}` instead of navigating to that specific directory beforehand.

.github/workflows/release.yml
```diff
@@ -30,7 +30,9 @@ jobs:
 
       - name: auto-dist-tag
         run: npx auto-dist-tag@1 --write
```
BobrImperator, line 32: Note: seems weird we use npx here

.github/workflows/release.yml
```diff
@@ -30,7 +30,9 @@ jobs:
...
+        working-directory: qunit-dom
```
BobrImperator, line 33: Here the working-directory makes sense to me, so I'd not change that.

Note; Although on another note we should consider setting up scripts for that in the root package.json.

package.json
```diff
@@ -1,111 +1,24 @@
...
-  },
-  "dependencies": {
-    "broccoli-funnel": "^3.0.8",
-    "broccoli-merge-trees": "^4.2.0",
-    "ember-cli-babel": "^7.26.10",
-    "ember-cli-version-checker": "^5.1.2"
+    "docs": "npm run build && npm exec documentation build qunit-dom/dist/qunit-dom.js --config documentation.yml -f md -o API.md",
+    "build": "cd qunit-dom && npm run build",
+    "release": "cd qunit-dom && npm run release",
+    "lint": "pnpm --filter '*' lint",
```
BobrImperator, line 7: `pnpm --filter '*'` -> `pnpm -r` should have the same effect.
It doesn't matter though, wouldn't change that.

qunit-dom/package.json
```diff
@@ -0,0 +1,109 @@
...
+  },
+  "dependencies": {
+    "broccoli-funnel": "^3.0.8",
+    "broccoli-merge-trees": "^4.2.0",
+    "ember-cli-babel": "^7.26.10",
+    "ember-cli-version-checker": "^5.1.2"
+  },
+  "devDependencies": {
+    "@ember/optional-features": "2.0.0",
+    "@ember/test-helpers": "2.9.4",
```
BobrImperator, line 49: Seems like the `qunit-dom` addon code has too many dependencies specified here.
This directory is no longer a host to the ember testing app, so basically all ember deps should be removed.

qunit-dom/package.json
```diff
@@ -0,0 +1,109 @@
...
+      "/node_modules/",
+      "/dist/"
+    ]
+  },
+  "dependencies": {
+    "broccoli-funnel": "^3.0.8",
+    "broccoli-merge-trees": "^4.2.0",
+    "ember-cli-babel": "^7.26.10",
+    "ember-cli-version-checker": "^5.1.2"
+  },
```
BobrImperator, line 46: V2 Addons should use @embroider/addon-dev package, right?

BobrImperator (commented):
Looks fine.
Some of the comments I added seem like they can be ignored since I made them before seeing https://github.com/mainmatter/qunit-dom/pull/2028.

For consistency across our other repos, I'd like to ask to put each of the packages to `/packages/` directory.
Other than that everything looks perfect :+1:

qunit-dom/package.json
NullVoxPopuli: Yes, but this isn't yet a v2 addon.

qunit-dom/package.json
NullVoxPopuli: These can be removed once v2 addon conversation happens

NullVoxPopuli:
note, I will be discarding all package.json / lockfile changes because having a chain of 3 PRs to manage over a long period of time _while_ upstream is changing so much is hard. (normally this wouldn't be the case, but the repo is getting restructured)

NullVoxPopuli:
Also, I'm going to start squashing all my commits in each PR.
The entirety of https://github.com/mainmatter/qunit-dom/pull/2028 is now in conflict.

.github/workflows/release.yml
NullVoxPopuli: done

.github/workflows/release.yml
NullVoxPopuli: that's how it is here: https://github.com/mainmatter/qunit-dom/blob/master/.github/workflows/release.yml#L32, on main

.github/workflows/ci.yml
NullVoxPopuli: done

package.json
```diff
@@ -1,113 +1,28 @@
...
-    ]
+  "engines": {
+    "node": ">= 16.*",
+    "yarn": "use pnpm",
+    "npm": "use pnpm"
   },
   "volta": {
-    "node": "18.18.0",
-    "pnpm": "7.33.6"
+    "node": "16.20.2",
```
BobrImperator, line 25: Noticed only now.
We should drop 16.x and run against 18 already - something to do after the V2 conversion.

.github/workflows/release.yml
BobrImperator: Yeah :+1: commented only as a note to myself.

PR #2028, opened 2023-09-23
Convert to V2 Addon

NullVoxPopuli:
Title is a bit of a lie, because this is now "just an npm package" no emberisms, allowing usage outside the ember ecosystem.

In technicality, "V2 Addons" are a migration process and spec for writing ember things as "just an npm package". Things like the addon-shim and other embroider-build/addon-blueprint (v2 addon blueprint) conventions are things that make traditional component-using addons easy for forwards and backwards compatibility. Since qunit-dom has _no_ emberisms, we don't need to bring all of those compatibility / emberisms along with us.

----------------

~~Blocked on: https://github.com/mainmatter/qunit-dom/pull/2027~~
(The diff here will look huge until the above is merged).

## Use this Diff

until the #2027 is merged.

https://github.com/NullVoxPopuli/qunit-dom/compare/v2-addon...actual-v2-addon

It's a _much_ smaller set of changes

image

Alternatively, this PR could be merged and we can ignore #2027

------

Actual changes here are 90+% removals.

However, I added a package that uses Vite to prove cross-compatibility with the wider ecosystem.

Without changing anything in qunit-dom, but _just_ removing all the emberisms, we get an error
image

This PR resolves that.
Evidence:
image

Of note:
- I had to remove the `ember-addon` keyword, because _with it_, we _must_ provide an addon-main, and for V2 addons that would be depending on `@embroider/addon-shim`, which I don't think we'd want to do.

- I was able to determine via the ember-try config that the min-supported ember-setup includes ember-qunit@v4, which precedes the ember-qunit where ember-auto-import@v2 is required: https://github.com/emberjs/ember-qunit/blob/main/addon/CHANGELOG.md#v600-2022-10-07 -- as a result, v2 conversion will be a breaking change, bumping qunit-dom to v3


package.json
```diff
@@ -17,12 +17,12 @@
     "prettier": "^2.8.8"
   },
   "engines": {
-    "node": ">= 16.*",
+    "node": ">= 18.*",
```
BobrImperator, line 20: We should allow 16 still. We just won't care whether it runs or not.

qunit-dom/package.json
```diff
@@ -38,34 +47,13 @@
       "/dist/"
     ]
   },
-  "dependencies": {
```
BobrImperator, line 41: We need to include addon-shim here https://github.com/embroider-build/embroider/blob/main/packages/addon-shim/README.md

package.json
NullVoxPopuli: node doesn't matter anymore, because qunit-dom is only relevant in browsers

qunit-dom/package.json
```diff
@@ -5,7 +5,6 @@
   "keywords": [
     "browser",
     "dom",
-    "ember-addon",
```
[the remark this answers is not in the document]
NullVoxPopuli: > Why'd we want to remove this keyword?

we can't have `ember-addon` and have a vanilla ESM package.
To use `ember-addon` currently requires:
- we use addon-main.cjs
- include `@embroider/addon-shim`

which is a lot more work for the build system than we need.

So, qunit-dom is now a vanilla npm package, no emberisms

qunit-dom/package.json
NullVoxPopuli: we do not, because this isn't an _ember addon_, it's a vanilla package that non-ember projects can use.

qunit-dom/package.json
```diff
@@ -14,16 +13,31 @@
   "repository": "https://github.com/simplabs/qunit-dom",
   "license": "MIT",
   "author": "simplabs GmbH",
-  "types": "dist/qunit-dom.d.ts",
-  "directories": {
-    "doc": "doc",
-    "test": "tests"
+  "type": "module",
```
NullVoxPopuli, line 16: a true module!

(something v2 addons aren't yet allowed to be)

qunit-dom/package.json
```diff
@@ -14,16 +13,31 @@
...
+  ],
   "scripts": {
-    "build": "rollup -c",
+    "build": "rollup -c ./rollup.config.mjs",
     "changelog": "lerna-changelog",
-    "lint": "eslint . --cache",
-    "lint:fix": "eslint . --fix",
+    "lint": "concurrently 'npm:lint:*(!fix)' --names 'lint:'",
+    "lint:fix": "concurrently 'npm:lint:*:fix' --names 'fix:'",
+    "lint:package": "publint",
```
NullVoxPopuli, line 37: I set up publint and are-the-types-wrong, because I couldn't debug my exports/typesVersions, turns out I had a typo :sweat_smile: 

good tools that are allowed once `type=module` is present in the package.json

package.json
BobrImperator: Though considering the other comments, it could be true if it were an ember-addon.

qunit-dom/package.json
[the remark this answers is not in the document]
NullVoxPopuli: nay, things start failing with the removal of `ember-addon`, specifically in embroider builds.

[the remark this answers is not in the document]
NullVoxPopuli: and ember-cli keys off this keyword as well.
which... _isn't good_.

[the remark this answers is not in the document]
NullVoxPopuli: when using `ember-addon`, we are required to have a v1-addon entrypoint. this is typically the `index.js` in v1 addons, or the `addon-main.cjs` for v2 addons. we don't need nor want either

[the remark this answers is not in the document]
NullVoxPopuli: oh, and we'd have to remove `type=module`, and remove publint / are-the-types-wrong as well, since embroider doesn't support type=module for ember-v2-addons yet.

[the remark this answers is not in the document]
BobrImperator: That's a bit unfortunate, makes sense though :sweat_smile:

PR #2031, opened 2023-09-27
Drop ember qunit less than v6

NullVoxPopuli:
In preparation for making the v2 conversion flow, starting in https://github.com/mainmatter/qunit-dom/pull/2027 easier, I'm extracting some of the more isolated changes.

For an overall plan for how v2 addon conversion works,
- I've written up a guide: https://gist.github.com/NullVoxPopuli/eafc7dad6547de5e730098498b829e1f
- high level conversion process: https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md
- at the end, qunit-dom will be using the v2-addon blueprint https://github.com/embroider-build/addon-blueprint (which @mansona helps maintain :muscle: )
- additional information for addon authors: https://github.com/embroider-build/embroider/blob/main/docs/addon-author-guide.md

Also, in part, to help with maintenance burden, supporting less is easier as well. ember-qunit is on v8 now, and v4 and v5 are _very old_ and have "magic" build-time behavior, which can't be transferred to the v2 addon world.

Up next will be upgrading the embroider deps, because < v3 isn't really supported :upside_down_face: 

-------------------

Outstanding questions for the maintainers:
- I've added a note to the README about installing older (aka, todays) qunit-dom, if folks have older ember-qunit -- with these changes it would still (probably) be possible to use with older ember-qunit, but all the automatic setup stuff would need to be handled by the consumer.

-------------------

Things to keep in mind: In my subsequent PRs to qunit-dom, there may be more breaking changes. I'll make sure each is called out in the PR that introduces them, if not extracted to a separate PR (which would be my preference, depending on the situation that calls for the change)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,125 +1,167 @@
-lockfileVersion: 5.4
-
-specifiers:
-  '@ember/optional-features': 2.0.0
-  '@ember/test-helpers': 2.9.4
-  '@embroider/compat': 0.48.1
-  '@embroider/core': 0.48.1
-  '@embroider/macros': 0.48.1
-  '@embroider/webpack': 0.48.1
-  '@types/jest': 27.5.2
-  '@types/qunit': 2.19.6
-  '@typescript-eslint/eslint-plugin': 5.62.0
-  '@typescript-eslint/parser': 5.62.0
-  babel-plugin-transform-es2015-modules-commonjs: 6.26.2
-  babel-preset-env: 1.7.0
-  broccoli-funnel: ^3.0.8
-  broccoli-merge-trees: ^4.2.0
-  documentation: 14.0.2
-  ember-auto-import: 2.6.3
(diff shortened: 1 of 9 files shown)


tests/dummy/config/ember-try.js
```diff
@@ -6,32 +6,18 @@ module.exports = async function () {
...
-            'ember-qunit': '^4.0.0',
             'ember-source': '~3.28.0',
-            '@ember/test-helpers': null,
-            qunit: null,
           },
         },
       },
       {
-        name: 'ember-qunit-5',
+        name: 'ember-lts-4.12',
```
NullVoxPopuli, line 17: I also added the latest v4 LTS here, since we want _some_ matrix in our matrix

pnpm-lock.yaml
NullVoxPopuli, line 1: looks like the lockfile is out of date

PR #2058, opened 2023-10-06
v3 RC: TS: Fix `Assert` types.

NullVoxPopuli:
at some point in the shuffle the `declare global {}` interface merge stopped working.

as I dug in to it, I noticed that the `declare global` was discarded for the iife build, but retained in the ES build.

as part of this process i put the es output in its own folder.
then disabled the declarations output for iife.

this is a non-breaking change

<details><summary>previously</summary>

As a result, I've split rollup's output files into sub-folders of `dist`, `es` and `iife`.

We don't _need_ to split both (iife could still be the root of `dist`, for example), but they do need to be separate from each other so they don't stomp on eachother's emitted files.

_right now_, here is how the output looks: 
image

a bunch of the .d.ts files are duplicated, but I was thinking about maybe cleaning those up from the iife output in a separate PR, leaving just the index.js, since we don't need separate types for the iife use case

</details>

### DIFF
--- a/packages/qunit-dom/lib/install.ts
+++ b/packages/qunit-dom/lib/install.ts
@@ -1,6 +1,12 @@
 import DOMAssertions from './assertions';
 import { getRootElement } from './root-element';
 
+declare global {
+  interface Assert {
+    dom(target?: string | Element | null, rootElement?: Element): DOMAssertions;
+  }
+}
+
 export default function (assert: Assert) {
   assert.dom = function (target?: string | Element | null, rootElement?: Element): DOMAssertions {
     if (!isValidRootElement(rootElement)) {
--- a/packages/qunit-dom/package.json
+++ b/packages/qunit-dom/package.json
@@ -16,16 +16,16 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/addon-test-support/qunit-dom-modules.d.ts",
-      "default": "./dist/addon-test-support/index.js"
+      "types": "./dist/es/qunit-dom.d.ts",
+      "default": "./dist/es/index.js"
     }
   },
   "typesVersions": {
     "*": {
-      "*": "dist/addon-test-support/qunit-dom-modules.d.ts"
+      "*": "dist/es/qunit-dom.d.ts"
     }
   },
-  "types": "dist/qunit-dom.d.ts",
+  "types": "dist/es/qunit-dom.d.ts",
   "files": [
     "dist"
   ],
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ This will attach the APIs to QUnit's `assert` object.
 
 ### Ember projects using &lt; v6.x of ember-qunit
 
-Install qunit-dom v2.0.0
+Install qunit-dom v3.0.0
 
 ### `<script>` Tag
 
@@ -69,7 +69,7 @@ Load `qunit-dom.js` *after* `qunit.js`:
 
 ```html
 <script src="https://unpkg.com/qunitjs/qunit/qunit.js"></script>
-<script src="https://unpkg.com/qunit-dom/dist/qunit-dom.js"></script>
+<script src="https://unpkg.com/qunit-dom/dist/iife/qunit-dom.js"></script>
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Load `qunit-dom.js` *after* `qunit.js`:
 
 ```html
 <script src="https://unpkg.com/qunitjs/qunit/qunit.js"></script>
-<script src="https://unpkg.com/qunit-dom/dist/iife/qunit-dom.js"></script>
+<script src="https://unpkg.com/qunit-dom/dist/qunit-dom.js"></script>
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ This will attach the APIs to QUnit's `assert` object.
 
 ### Ember projects using &lt; v6.x of ember-qunit
 
-Install qunit-dom v3.0.0
+Install qunit-dom v2.0.0
 
 ### `<script>` Tag
 
--- a/packages/qunit-dom/lib/qunit-dom.ts
+++ b/packages/qunit-dom/lib/qunit-dom.ts
@@ -1,14 +1,6 @@
 /* global QUnit */
-
-import type DOMAssertions from './assertions';
 import install from './install';
 
 export { setup } from './qunit-dom-modules';
-
-declare global {
-  interface Assert {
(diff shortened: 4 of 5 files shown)


packages/qunit-dom/lib/install.ts
```diff
@@ -1,6 +1,12 @@
 import DOMAssertions from './assertions';
 import { getRootElement } from './root-element';
 
+declare global {
```
NullVoxPopuli, line 4: I wanted to move this to a common location that was imported by both the IIFE and ES entrypoints.

Additionally, given the file name, install, I thought it makes semantic sense to live here as well, since we're installing the `dom` property on `Assert`

BobrImperator:
> a bunch of the .d.ts files are duplicated, but I was thinking about maybe cleaning those up from the iife output in a
> separate PR, leaving just the index.js, since we don't need separate types for the iife use case

I'd lean towards that personally. It seems like it's possible to tell the plugin to not emit the `d.ts` files for the `iife` build.
With that it seems like we could just leave the dist as it was before?

https://www.npmjs.com/package/rollup-plugin-typescript2#declarations

NullVoxPopuli:
It seems like that'll require a second tsconfig.json (which I can/will do)

packages/qunit-dom/package.json
```diff
@@ -16,16 +16,16 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/addon-test-support/qunit-dom-modules.d.ts",
-      "default": "./dist/addon-test-support/index.js"
+      "types": "./dist/es/qunit-dom.d.ts",
+      "default": "./dist/es/index.js"
```
BobrImperator, line 20: Do we need this paths change too?
I think we should not change them if we don't need to.
Especially since the issue they originally attempted to solve was fixed with the configuration for 'iife' output.

NullVoxPopuli: I can split it out if you want, but I think it helps immensely with understanding

BobrImperator: Overall I'm currently struggling to begin treating qunit-dom as a JS lib rather than an addon.
I agree with the changes though, guess it makes sense to rename that, since as you've mentioned previously there's no "emberism" anymore.

I'd like to ask is an `es` directory a general practice in the JS land?
I think it makes sense but was wondering if this is how it's usually done.

At this point, I believe we should also implement the convention to add an `.esm` extension to the file instead so
`/dist/es/index.js` -> `/dist/index.esm.js`
Though I've no real hard opinion on this, just wondering what'd be the most common and appropriate approach here. I've looked at outputs of a few libraries and interestingly each one did it differently :joy: 
One common approach that some of them have is `.esm.js` naming convention and ideally keep things flat.

I don't know which is best and what'd be the pros and cons of each one.
Personally I think I like what `popper.js` does best; they basically say "here's the default way to consume the package" while also including the same assets in a specific directory as a failsafe.
And then also in case of `photoswipe` types are also in a directory of their own.

At the end of the day, users will by default use whatever we declare in the package.json but I'd like to decide here what approach should we take and why.

What do you think?

Examples:
https://unpkg.com/browse/vue@3.3.4/dist/
- Vue seems to have dropped any kind of additional nesting and all files are kept in the `/dist/` and distinguishes build types based on extensions e.g. `.esm.js`, `.cjs.js`

https://unpkg.com/browse/react-dom@18.2.0/
- might not be exactly applicable as it's AFAIK a javascript that runs only in node environment, still leaving this here

https://www.unpkg.com/browse/photoswipe@5.4.2/dist/
- seems to have "default" `esm` export in the `/dist/` directly with additional file extension `.esm.js` 

https://unpkg.com/browse/popper.js@1.12.8/dist/
- seems to have both "default" `esm` export in the `/dist/` and also the same thing copied to `/dist/esm/`

NullVoxPopuli: > At this point, I believe we should also implement the convention to add an .esm extension to the file instead so

there's no need because the package.json declares `type=module`, so "it's known" that all `.js` files are ESM.

the IIFE is the odd one out, because IIFEs have no place in node/bundler land. They only make sense in old school CDN usage. And even, then, we can use ESM in CDNs now, for example:
```html
Note how qunit-dom has no special treatment to the path (other than targeting the pre-release version), because we end up hitting the "exports" entry in the package.json, which ESM-based CDNs understand and respect 🥳

<script type="importmap">
  {
    "imports": {
      "qunit": "https://esm.sh/qunit",
      "qunit-dom": "https://esm.sh/qunit-dom@3.0.0-rc.1"
    }
  }
</script>
<link rel="stylesheet" href="https://esm.sh/qunit@2.20.0/qunit/qunit.css">
<script type="module">
  import { setup } from 'qunit-dom';

  setup(QUnit.assert);

</script>
```

So, _technically_ the legacy IIFE build isn't needed at all, folks wanted to use native ESM in browsers via CDN.

Here is a demonstration of how we can use `qunit-dom` on jsbin:
https://jsbin.com/fipayiy/1/edit?html,output
image


> I'd like to ask is an es directory a general practice in the JS land?

I wouldn't say there are general practices, but directories per format _are_ quite common, and all the glimmer packages do it as well.
image
(tho, AMD really has no place in modern packages)


> I've looked at outputs of a few libraries and interestingly each one did it differently 😂

yeah, I mean, ultimately, it's up to each project to do what works best for them.
The package.json and exports / imports / ESM rules are what we all abide by, and beyond that, anything is fair game.

> "here's the default way to consume the package" while also including the same assets in a specific directory as a failsafe.

the whole story of how to consume a package is defined by the `exports`.
for example, if we wanted to also support commonjs (which doesn't make sense for a browser library, such as qunit-dom), we'd do:
```
"exports": {
  "*": {
    "types": "... types must be first ...",
    "require": " path for cjs-only file ",
    "import": " path for esm-only file"
  }
}
```
when using `exports.*.default`, you say all types of resolving use "this file" (the default) -- so it ends up working for cjs and esm.

`main` / `module` / `browser` fields in the package.json aren't really used anymore for browser packages.
`main` still has some usage in node libraries though.



> Vue seems to have dropped any kind of additional nesting and all files are kept in the /dist/ and distinguishes build types based on extensions

this is easiest achieved when the build tools allow for it.
since we're essentially running rollup twice, separate target directories has

BobrImperator: That's a blog post entry right there :open_mouth: 
Thanks a lot for this!

I thought I've had a seemingly innocent question going along these lines "How qunit-dom's build artifacts map to the general state of JS land?" and I received this gold in response :pray:

BobrImperator:
Released as RC2 https://www.npmjs.com/package/qunit-dom/v/3.0.0-rc.2

PR #2087, opened 2023-12-09
feature: support IDOMElementDescriptors

bendemboski:
Implement support for [RFC #726](https://github.com/emberjs/rfcs/pull/726)

### DIFF
--- a/packages/qunit-dom/lib/__tests__/does-not-have-tagname.ts
+++ b/packages/qunit-dom/lib/__tests__/does-not-have-tagname.ts
@@ -63,7 +63,7 @@ describe('assert.dom(...).doesNotHaveTagName()', () => {
 
       expect(assert.results).toEqual([
         {
-          message: 'Element <unknown> should exist',
+          message: 'Element <not found> should exist',
           result: false,
         },
       ]);
--- a/packages/qunit-dom/lib/__tests__/has-tagname.ts
+++ b/packages/qunit-dom/lib/__tests__/has-tagname.ts
@@ -63,7 +63,7 @@ describe('assert.dom(...).hasTagName()', () => {
 
       expect(assert.results).toEqual([
         {
-          message: 'Element <unknown> should exist',
+          message: 'Element <not found> should exist',
           result: false,
         },
       ]);
--- a/packages/qunit-dom/tsconfig.json
+++ b/packages/qunit-dom/tsconfig.json
@@ -6,11 +6,12 @@
     "verbatimModuleSyntax": true,
     // We build to ESM, so we need to tell TS that we build to ESM
     // otherwise TS uses fake ESM (you use imports, but it assumes the compilation is to CJS)
-    // These are also required when using 
-    // verbatimModuleSyntax, because it's an 
+    // These are also required when using
+    // verbatimModuleSyntax, because it's an
     // ESM-only feature
     "target": "esnext",
     "module": "esnext",
+    "moduleResolution": "node",
 
     "lib": [
       "ES6",
--- a/packages/qunit-dom/lib/__tests__/descriptor.ts
+++ b/packages/qunit-dom/lib/__tests__/descriptor.ts
@@ -1,0 +1,119 @@
+import { describe, beforeEach, test, expect } from 'vitest';
+
+import createDescriptor from '../descriptor';
+import TestAssertions from '../helpers/test-assertions';
+import {
+  resolveDOMElement,
+  resolveDOMElements,
+  resolveDescription,
+  createDescriptor as createDescriptorBase,
+} from 'dom-element-descriptors';
+
+describe('createQuery()', () => {
+  beforeEach(() => {
+    document.body.innerHTML =
+      '<div class="single"></div><div class="multiple"></div><div class="multiple"></div>';
+  });
+
+  test('it works with a selector', () => {
+    let expected = Array.from(document.querySelectorAll('.multiple'));
+    let descriptor = createDescriptor('.multiple', document);
+
+    expect(resolveDOMElement(descriptor)).toEqual(expected[0]);
+    expect(resolveDOMElements(descriptor)).toEqual(expected);
+    expect(resolveDescription(descriptor)).toEqual('.multiple');
+  });
+
+  test('selector respects the root element', () => {
+    document.body.innerHTML =
+      '<div class="multiple"><div class="root"><div class="multiple"></div></div>';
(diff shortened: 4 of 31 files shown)


bendemboski:
This PR does change existing error messages, as can be seen from the tests, which isn't an explicit goal of this change. Here's the explanation.

If an assertion couldn't find its element (whether because of a non-matching selector or because it was passed `null`), previously the error messaging would be inconsistent between its usage of `<not found>` vs. `<unknown>`. The code that produced those messages was refactored and is now driven by the dom element descriptor description functionality, so it would require extra code to restore the inconsistent behavior and keep the error message the same as it was before this PR.

BobrImperator (commented):
`qunit-dom` attempts to stay lean and it also promotes a way of testing without using page-objects. I understand the need and how its useful for some use cases.

I'd like us to look for a way for `dom-element-descriptors` to depend on `qunit-dom` instead and not the other way .around.
It seems like `qunit-dom` at the very minimum would need to provide a way to extend the element lookup and descriptions if they're required. and the `assert.dom` method.

Then what users of `dom-element-descriptors` would need to do is swap out `qunit-dom` setup with `dom-element-descriptors` setup. 
```js
// tests/test-helper.js
import { setup } from 'dom-element-descriptors';

//...
setup(QUnit.assert);
setApplication(Application.create(config.APP));
start();
//...
```

Some pseudo code for potential qunit-dom extension could be:

```js
import { setup, DOMHandler } from 'qunit-dom'

// A class or anything that implements similar interface
class DOMElementDescriptorHandler implements DOMHandler {
  findElement(selector, rootElement)
  findElements(selector, rootElement)
  description(selector)
  resultMessage()
}

setup(QUnit.assert, [new DOMElementDescriptorHandler()])
```

Draft https://github.com/mainmatter/qunit-dom/pull/2101

BobrImperator:
Drafted https://github.com/mainmatter/qunit-dom/pull/2101 as per feedback, would this help `dom-element-descriptors` to create a "plugin" for qunit-dom?

packages/qunit-dom/lib/__tests__/does-not-have-tagname.ts
```diff
@@ -63,7 +63,7 @@ describe('assert.dom(...).doesNotHaveTagName()', () => {
 
       expect(assert.results).toEqual([
         {
-          message: 'Element <unknown> should exist',
+          message: 'Element <not found> should exist',
```
BobrImperator, line 66: Please revert the message changes back to <unknown>.
If it's something really needed, I'd prefer it to be a separate PR.

```suggestion
          message: 'Element <unknown> should exist',
```

BobrImperator (changes requested):
Please revert the message changes and rebase the PR.

packages/qunit-dom/lib/__tests__/has-tagname.ts
```diff
@@ -63,7 +63,7 @@ describe('assert.dom(...).hasTagName()', () => {
 
       expect(assert.results).toEqual([
         {
-          message: 'Element <unknown> should exist',
```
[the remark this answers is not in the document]
bendemboski: Okay, I reverted all changes to existing tests, so the messages are the same as in `master`. To make this happen, I had to introduce a bit of screwiness to the code:

* [selectedBy](https://github.com/mainmatter/qunit-dom/pull/2087/files#diff-d38c97d68f44c96157e0203728650e0b5547fe8a550244635badd83b8b35ded3R1459-R1466) has to customize its message based on whether an element was matched or not, to mirror [the previous logic](https://github.com/mainmatter/qunit-dom/blob/a420f27f323904a1073be837f3f1909549eca095/packages/qunit-dom/lib/assertions.ts#L1224). This seems a little odd because in my mind, an argument of `null` usually means that the caller tried to pass an `Element`, but something like `querySelector()` didn't match anything and ended up evaluating to `null` and that was passing into `assert.dom()`. So I would expect to see `passed` in this case (tried to pass an `Element` that happened to be `null`), rather than `selected by null` which kinda implies a selector (string) containing `null` was passed.
* `exists()` and `isVisible()` had to do something screwy ([here](https://github.com/mainmatter/qunit-dom/pull/2087/files#diff-95990b5c77186fc52f6d6ba30f6aa46e1919752a372ea0a2202e2698935cde03R40-R42) and [here](https://github.com/mainmatter/qunit-dom/pull/2087/files#diff-1448fb716f5de96ca01505b59eed5840029a6ce50eab4852599f82c2d28a180cR41-R43)) to mirror the fact previously most assertions used `findTargetElement()` which would error with <[unknown](https://github.com/mainmatter/qunit-dom/blob/a420f27f323904a1073be837f3f1909549eca095/packages/qunit-dom/lib/assertions.ts#L1413)> if passed `null`, while `exists()` and `isVisible()` used [targetDescription](https://github.com/mainmatter/qunit-dom/blob/a420f27f323904a1073be837f3f1909549eca095/packages/qunit-dom/lib/assertions/exists.ts#L16) which called `elementToString()` which would error with <[not found](https://github.com/mainmatter/qunit-dom/blob/a420f27f323904a1073be837f3f1909549eca095/packages/qunit-dom/lib/helpers/element-to-string.ts#L4)>.

I'm fine leaving it as-is with these oddities, or creating a follow-on PR that removes them and changes some of the messaging like this PR originally did -- let me know what you prefer.

bendemboski:
@BobrImperator this is rebased and, pending your decision on the assertion messages change or any other feedback you might have, ready to merge!

packages/qunit-dom/tsconfig.json
```diff
@@ -7,11 +7,12 @@
...
     // We build to ESM, so we need to tell TS that we build to ESM
     // otherwise TS uses fake ESM (you use imports, but it assumes the compilation is to CJS)
-    // These are also required when using 
-    // verbatimModuleSyntax, because it's an 
+    // These are also required when using
+    // verbatimModuleSyntax, because it's an
     // ESM-only feature
     "target": "esnext",
     "module": "esnext",
+    "moduleResolution": "node",
```
BobrImperator, line 15: Let's use `"bundler"` option here instead of `"node"` since it seems like rollup already requires it  in v4 when typescript is involved 

https://vitejs.dev/guide/migration#rollup-4

BobrImperator (approved):

BobrImperator:
https://github.com/mainmatter/qunit-dom/releases/tag/v3.2.0

PR #2143, opened 2024-10-09
feature: use `Element.computedStyleMap()`

BarryThePenguin:
Uses [`Element.computedStyleMap()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/computedStyleMap) to evaluate element styles when it is available. Firefox is the only browser that hasn't implemented CSS Typed OM https://ishoudinireadyyet.com/

I was running into issues where `hasStyle()` was returning outdated values. Specifically when an element had a css `height: 100px` and JavaScript that updated it with an inline style of `height: 85%`, it would return the old `height: 100px` value

### DIFF
--- a/packages/qunit-dom/lib/assertions.ts
+++ b/packages/qunit-dom/lib/assertions.ts
@@ -10,6 +10,7 @@ import isVisible from './assertions/is-visible.js';
 import isDisabled from './assertions/is-disabled.js';
 import matchesSelector from './assertions/matches-selector.js';
 import collapseWhitespace from './helpers/collapse-whitespace.js';
+import getComputedStyle from './helpers/get-computed-style.js';
 import {
   type IDOMElementDescriptor,
   resolveDOMElement,
@@ -28,10 +29,6 @@ export interface AssertionResult {
 export interface ExistsOptions {
   count: number;
 }
-
-type CSSStyleDeclarationProperty = keyof CSSStyleDeclaration;
-
-type ActualCSSStyleDeclaration = Partial<Record<CSSStyleDeclarationProperty, unknown>>;
 
 /**
  * @namespace
@@ -832,8 +829,8 @@ export default class DOMAssertions {
     let element = this.findTargetElement();
     if (!element) return this;
 
-    let computedStyle = window.getComputedStyle(element, selector);
-    let expectedProperties = Object.keys(expected) as CSSStyleDeclarationProperty[];
+    let computedStyle = getComputedStyle(element, selector);
+    let expectedProperties = Object.keys(expected);
     if (expectedProperties.length <= 0) {
       throw new TypeError(
         `Missing style expectations. There must be at least one style property in the passed in expectation object.`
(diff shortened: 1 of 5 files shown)


packages/qunit-dom/package.json
```diff
@@ -41,6 +41,7 @@
   "devDependencies": {
     "@arethetypeswrong/cli": "0.15.4",
     "@types/qunit": "2.19.10",
+    "@typescript/lib-dom": "npm:@types/web@0.0.169",
```
BarryThePenguin, line 44: https://www.npmjs.com/package/@types/web for the new Web APIs

packages/qunit-dom/lib/assertions.ts
```diff
@@ -29,10 +30,6 @@ export interface ExistsOptions {
   count: number;
 }
 
-type CSSStyleDeclarationProperty = keyof CSSStyleDeclaration;
-
-type ActualCSSStyleDeclaration = Partial<Record<CSSStyleDeclarationProperty, unknown>>;
```
BarryThePenguin, line 34: I don't think these types provide much value as they are only used internally, with a type assertion..

I think these types are better represented as a plain record, eg; `Record<string, string>`

packages/qunit-dom/lib/assertions.ts
```diff
@@ -832,26 +829,20 @@ export default class DOMAssertions {
     let element = this.findTargetElement();
     if (!element) return this;
 
-    let computedStyle = window.getComputedStyle(element, selector);
-    let expectedProperties = Object.keys(expected) as CSSStyleDeclarationProperty[];
+    let computedStyle = getComputedStyle(element, selector);
+    let expectedProperties = Object.keys(expected);
```
BarryThePenguin, line 833: 🗑️ the type assertion

packages/qunit-dom/lib/assertions.ts
```diff
@@ -832,26 +829,20 @@ export default class DOMAssertions {
...
       throw new TypeError(
         `Missing style expectations. There must be at least one style property in the passed in expectation object.`
       );
     }
 
     let result = expectedProperties.every(
-      property =>
-        (computedStyle.getPropertyValue(property.toString()) || computedStyle[property]) ===
-        expected[property]
+      property => computedStyle[property] === expected[property]
```
BarryThePenguin, line 841: `camelCase` and `kebab-case` properties are handled by `getComputedStyle()`

packages/qunit-dom/lib/helpers/get-computed-style.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        if (typeof property === 'string' && computedStyleMap) {
+          value = computedStyleMap.get(property)?.toString();
+        }
+
+        return (
+          value ||
+          computedStyle.getPropertyValue(property.toString()) ||
+          Reflect.get(computedStyle, property)
+        );
```
BarryThePenguin, line 24: Maintains the previous logic of

```js
computedStyle.getPropertyValue(property.toString()) || computedStyle[property]
```

PR #2147, opened 2024-11-10
feature: add HTML matching methods

### LINKED ISSUE #2025: Add `hasHtml` to assert an element has specific HTML
When testing whether dynamic content is rendered correctly as HTML, it would be useful to have a `hasHtml` assertion.

This would be the same as `hasText` except that instead of testing against `textContent`, it would test against `innerHTML.trim()`.

The workaround for now is use `assert.equal`. For instance:

```
const html =  'Where you put your <b>emphasis</b> is up to you.',

// cause this HTML to be rendered

assert.equal(document.querySelector('.message').innerHTML.trim(), html);
```

However, having an `hasHtml` built-in would be more idiomatic:

```
assert.dom('.message').hasHtml(html);
```

CvX:
Examples:

```js
assert.dom('h1').hasHtml('A <b>great</b> thing');
assert.dom('section').doesNotHaveHtml('<b>fantastic</b>');

assert.dom('#title').includesHtml('<b>nice</b>');
assert.dom('#title').doesNotIncludeHtml('<i>nope</i>');
```

closes #2025

### DIFF
--- a/API.md
+++ b/API.md
@@ -115,6 +115,18 @@
     *   [doesNotHaveTagName][111]
         *   [Parameters][112]
         *   [Examples][113]
+    *   [hasHtml][114]
+        *   [Parameters][115]
+        *   [Examples][116]
+    *   [doesNotHaveHtml][117]
+        *   [Parameters][118]
+        *   [Examples][119]
+    *   [includesHtml][120]
+        *   [Parameters][121]
+        *   [Examples][122]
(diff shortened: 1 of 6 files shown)


BobrImperator (approved):
This looks great! Thanks heaps

MrChocolatine:
These are really cool additions. Thanks.

PR #2153, opened 2024-12-01
feature: add includesValue/doesNotIncludeValue

CvX:
`hasValue(regex)` works for simple strings, but if you need to check whether an input/textarea includes a more complex string like markdown and/or url, converting it to a regular expression might not be feasible.


BobrImperator (approved):
Looks good, thanks!

PR #2185, opened 2025-08-19
chore(qunit-dom): expose private install types

BobrImperator:
`ember-data` found a need to import the Assert interface for it's internal Diagnostic module.
This exposes the types so it's not necessary to patch the dependency directly 

https://github.com/emberjs/data/pull/10122


bartocc:
I might be wrong, but I think this PR changed the way `assert.dom(undefined)` behaves. Under 3.4.0, it was equivalent to `assert.dom()` while now, it errors

I wonder if this change should have been clearer in the CHANGELOG, and maybe justify a major bump 🤔

PR #2201, opened 2026-04-14
Run create-release-plan for updates

evoactivity:


evoactivity:
It seems updating this caused some transitive dependencies to update (JSDOM/ nwsapi) changing the way some tests work.

1) `window.computedStyle` returns colours as `rgb(x, x, x)` values (this matches the real DOM).
2) uppercase selectors fail so I skipped that test.

